PR #24, opened 2015-01-26
document how to install new javascript, css, sass dependencies with new wiredep configuration

andidev:
fix jhipster/generator-jhipster#1030


jdubois:
As this will be published immediately after merge, I'll merge this just after the next release (so documentation is on the same version as the generator)

andidev:
Sounds like a good idea, when will the next release be actually?
Feel free on commenting on the docs.

jdubois:
It's ready when I have time!
Probably this week, depending on what I do on Wednesday and Thursday.

PR #25, opened 2015-01-30
Change wording in intro

dsyer:
Spring Security works just fine with a SPA (and that's why jhipster is using it). It would be better to say so.

### DIFF
--- a/security.html
+++ b/security.html
@@ -9,7 +9,7 @@ sitemap:
 <h1><i class="fa fa-lock"></i> Securing your application</h1>
 
 <p>
-    Spring Security does not work well with Single Web Page Applications, like JHipster, as it implies you have some specific login/logout/error pages. We have modified Spring Security in order to have AngularJS views work correctly, and of course we generate all those views for you.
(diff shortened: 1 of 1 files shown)


jdubois:
Maybe this has changed, but I was referring to the 3 Ajax*Handler we had to code and configure:

https://github.com/jhipster/jhipster-sample-app/tree/master/src/main/java/com/mycompany/myapp/security

At least 1 year ago they were required.

dsyer:
I see. I don't really think they are mandatory (see my recent blog series where I manage without, for example). But that's not really the point - the point is that the wording was strange: if you say that Spring Security is not suitable, the reader expect you to say that you are using something else, whereas you manage to use it without too much fuss, and I'm pretty sure there are benefits in doing so. I was only trying to clarify that a bit.

Did you raise any JIRA tickets in Spring Security for the features that you think are missing? There's this one: https://jira.spring.io/browse/SEC-2803, but maybe you know of more?

jdubois:
Of course I'll merge your PR, but I will add it's "Ajax" pages, because that's where the issue comes from, and I think it still exists.

I will check what I can send to Spring Security, but I think I already filled Jiras for those, probably 3 or 4 years ago (this is in fact old code from former projects of mine!)

PR #26, opened 2015-01-30
Added a new chapter about how to customise Bootstrap

gmarziou:
Added a new chapter about how to customise Bootstrap in JHipster context.
I added a css just for trying to mimic the Bootstrap logo but I could make it simpler as I did in the menu.

Fix #1063

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -17,6 +17,9 @@
     <!-- Bootstrap Core CSS -->
     <link href="{{ site.url }}/css/bootstrap.min.css" rel="stylesheet">
 
+    <!-- Bootstrap Docs CSS subset to get icon -->
+    <link href="{{ site.url }}/css/bootstrap-docs.css" rel="stylesheet">
+
     <!-- MetisMenu CSS -->
     <link href="{{ site.url }}/css/plugins/metisMenu/metisMenu.min.css" rel="stylesheet">
 
@@ -89,6 +92,7 @@
             				<li {% if '/creating_a_service.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/creating_a_service.html"><i class="fa fa-bolt"></i> Creating a service</a></li>
                             <li {% if '/installing_new_languages.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/installing_new_languages.html"><i class="fa fa-flag"></i> Installing new languages</a></li>
                             <li {% if '/using_angularjs.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/using_angularjs.html"><i class="fa fa-university"></i> Using AngularJS</a></li>
+                            <li {% if '/using_angularjs.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/customising_bootstrap.html"><i class="fa fa-border">B</i> Customising Bootstrap</a></li>
                             <li {% if '/security.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/security.html"><i class="fa fa-lock"></i> Securing your app</a></li>
                             <li {% if '/profiles.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/profiles.html"><i class="fa fa-users"></i> Managing profiles</a></li>
             				<li {% if '/development.html' == page.url %}class="active"{% endif %}><a href="{{ site.url }}/development.html"><i class="fa fa-code"></i> Using in development</a></li>
--- a/css/bootstrap-docs.css
+++ b/css/bootstrap-docs.css
@@ -1,0 +1,19 @@
+/* few CSS classes extracted from Boostrap docs to be able to render the Bootstrap icon */
+
+.bs-docs-booticon {
+    color: #FFF;
+	background-color: #6F5499;
+    display: inline-block;
+    text-align: center;
+    border-radius: 15%;
+}
+
+.bs-docs-booticon-lg {
+    width: 60px;
+    height: 60px;
+    font-size: 54px;
+}
+
+.bs-docs-booticon-outline {
+    border: 1px solid #CDBFE3;
+}
--- a/customising_bootstrap.md
+++ b/customising_bootstrap.md
@@ -1,0 +1,75 @@
+---
+layout: default
+title: Customising Bootstrap
+sitemap:
+priority: 0.7
+lastmod: 2015-01-30T18:40:00-00:00
+---
+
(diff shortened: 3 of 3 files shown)


gmarziou:
I just tested the import on a fresh project and both syntax works:

  // bower:scss
  @import "bootstrap-sass/assets/stylesheets/_bootstrap.scss";
  // endbower

is equivalent to:

  // bower:scss
  @import "bootstrap-sass/assets/stylesheets/bootstrap";
  // endbower

The second syntax is more idiomatic, it's the way _bootstrap.scss imports other partials.
Wiredep generates the first one, this is annoying as I have to explain it in the doc.

Official doc on partails: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#partials

jdubois:
@gmarziou I've been merging this locally 2 hours ago and I'm going to push it -> if you want to change things can you do another PR?

gmarziou:
I found this issue https://github.com/yeoman/generator-gulp-webapp/issues/85 which makes me think that wiredep can create problems with Compass depending on how the imported component is packaged.

Also, wiredep can only inserts default scss so in the case of Bootstrap it imports all Bootstrap which is not what I want, so in this case I would need to exclude Bootstrap from the files imported by wiredep.

My recommandation would be to exclude bootstrap-sass from wiredep, so the main.scss template would look like:

    $icon-font-path: "../../bower_components/bootstrap-sass/assets/fonts/bootstrap/";
    @import "bootstrap-sass/assets/stylesheets/bootstrap";
    // bower:scss
   // endbower


Another issue with te way wioredep process scss file is that it does not know what is the right order of the imports becasue yes it matters for SASS.

So globally, I think wiredep is not well sujited for SASS files.

gmarziou:
I'll do a new PR but I think we shoudl re-consider using wiredep for SASS files, I suspect it could become a source of issues with  users installing components embedding SASS files expectign Bootstrap-sass file to be inclmuded first.

jdubois:
OK, I trust you on this, as I don't use Compass and have no idea how you should use it :-)

gmarziou:
So let's wait what others think, in particular @andidev

gmarziou:
At least if we keep wiredep for SASS files, we should document how to exclude some particular directories in the Gruntfile

gmarziou:
I can see that some components (angular-i18n and swagger-ui) are already excluded from wiredep task in Gruntfile.js

andidev:
I don't think that using wiredep for sass i a problem. I'm gonna try to explain how the wiredep works in detail and then it will make sense:

When the wiredep task is used for injecting say for example "Bootstrap Sass" it starts with checking if the project has dependencies. This is done by looking in the "Bootstrap Sass" [bowed.json](https://github.com/twbs/bootstrap-sass/blob/v3.3.3/bower.json) configuration file.

More exactly it looks at the [bower.json dependencies property](https://github.com/twbs/bootstrap-sass/blob/v3.3.3/bower.json#L37-L39) which looks like this
```json
  "dependencies": {
    "jquery": ">= 1.9.0"
  }
```

Since there exists a dependency, bower will make sure that this dependency is injected before the "Bootstrap Sass" project files. So if the owner a plugin has configured their bower.json file correctly, the dependecies should be injected in a correct order.

After the dependencies has been injected it starts with injecting the files that are specified in the [bower.json main property](https://github.com/twbs/bootstrap-sass/blob/v3.3.3/bower.json#L12-L20). For the Bootstrap Sass the main property currently looks like this
```json
  "main": [
    "assets/stylesheets/_bootstrap.scss",
    "assets/fonts/bootstrap/glyphicons-halflings-regular.eot",
    "assets/fonts/bootstrap/glyphicons-halflings-regular.svg",
    "assets/fonts/bootstrap/glyphicons-halflings-regular.ttf",
    "assets/fonts/bootstrap/glyphicons-halflings-regular.woff",
    "assets/fonts/bootstrap/glyphicons-halflings-regular.woff2",
    "assets/javascripts/bootstrap.js"
  ],
```
Bower will start injecting files from the top to the bottom as they are defined in the array.

These injections will however only be done for a filetype if the corresponding codeblock comment is encountered in the file that is configured for injection.

In our Gruntfile.js the main.scss file is configured for wiredep injection 
```javascript
        wiredep: {
            app: {
                src: ['src/main/webapp/index.html', 'src/main/scss/main.scss'],
                exclude: [/angular-i18n/, /swagger-ui/],
                ignorePath: /\.\.\/webapp\/bower_components\// // remove ../webapp/bower_components/ from paths of injected sass files
            },
```

and our main.scss file contains the following codeblock comment telling wiredep to inject sass files
```
// bower:scss
// endbower
```

So the conclusion is this: As long as the developers of the installed bower components has configured their dependencies correctly we should not have a problem with the dependency order. The issue [yeoman/generator-gulp-webapp#85](yeoman/generator-gulp-webapp#85 ) in the Gulp Webapp Generator is actually derived from that Bootstrap configured their bower.json in an incorrect way. This is how the [bower.json](https://github.com/twbs/bootstrap-sass/blob/v3.1.1%2B1/bower.json) looked like at the moment when the issue was encountered. As you can see the main property incorrectly contains all the bootstrap sass partials in alfabetic order.

In the case when customizing bootstrap by using a _custom-bootstrap.scss file instead of the wiredepped _bootstrap.scss. The wiredepped _bootstrap.scss file needs to be excluded from being injected by wiredep just as you say @gmarziou. As you also mention the way to do this is by adding it to the exclude list of the wiredep connfiguration e.g.
```
javascript
        wiredep: {
            app: {
                src: ['src/main/webapp/index.html', 'src/main/scss/main.scss'],
                exclude: [/angular-i18n/, /swagger-ui/, /_bootstrap.scss /]],
                ignorePath: /\.\.\/webapp\/bower_components\// // remove ../webapp/bower_components/ from paths of injected sass files
            },
```

But note that it is important **not** to remove the 
```
// bower:scss
// endbower
```
codeblock comment from the main.scss file since we still want other bower components to wiredep themselves there.

It is also important not to put any other sass imports inside this codeblock manually because the grunt wiredep task will remove them and replace them with the bower components it is configured to inject.
So the _custom-bootstrap.scss file needs to be added before the bower codeblock comment. Or of cource after if it has some dependency on any injected bower component.

@gmarziou I now see that this is actually exactly what you wrote. Though if I understand you correctly you want the remove the bootstrap from being wiredeped. Is'nt it better to have it wiredepped as default and explain that it has to be excluded in the docs about customizing bootstrap? 

I think this looong answer should cover it. Hope this makes sense =)

gmarziou:
Thanks @andidev for taking the time to explain this into details, I think my new PR fits well with your explainations.
Please have a look.

https://github.com/jhipster/jhipster.github.io/pull/28

PR #30, opened 2015-02-08
Added instructions to install either grunt or gulp.

gmarziou:
Fix jhipster/generator-jhipster#1075

### DIFF
--- a/installation.html
+++ b/installation.html
@@ -22,12 +22,12 @@ We provide 2 ways of working with JHipster:
 
 <ol>
     <li>Install Java from <a href="http://www.oracle.com/technetwork/java/javase/downloads/index.html" target="_blank">the Oracle website</a>.</li>
-    <li>Depending on your preferences, install either <a href="http://maven.apache.org/" target="_blank">Maven</a> or <a href="http://www.gradle.org/" target="_blank">Gradle</a>.</li>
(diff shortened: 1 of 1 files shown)


jdubois:
Yes this need to be updated, but:

- I would like the correct wording "Grunt" and "Gulp.js", and have the links to both tools (like for Maven and Gradle)
- And I would like to clearly say our default is Grunt -> most Java devs don't know either of those tools, and I don't want them to be lost on this part (btw, we must say the same thing for Maven)

gmarziou:
OK, I'll make these changes.

gmarziou:
Each step description is quite concise but if you want to target beginners in NodeJS then maybe there should be little more like a short verification after each installation.

Something like: install bower, run bower -v, maybe it could be visually not intrusive maybe using columns

| Installation  | Verification |
|-----------------|-----------------|
| Install Bower `npm install -g bower` | `bower -v` should return something like `1.3.12` |

Or maybe some collaspible section under each step.

jdubois:
I'm merging your PR but yes should be improved... This is possibly our most important page, and it's not good enough -> both of your proposal are good. Another good documentation is https://docs.npmjs.com/

gmarziou:
I'll try to come up with something., maybe Compass should be listed.

PR #31, opened 2015-02-10
Mention the gradle  wrapper instead of installing gradle

atomfrede:
Not sure about the wording though...


jdubois:
OK, I'm rewording this...

PR #32, opened 2015-02-13
Added documentation about Graddle wrapper

gmarziou:


development.html
```diff
@@ -51,18 +51,30 @@
...
     <code>
         gradle bootRun
     </code>
 </p>
+<p>
+    If you haven't installed Gradle, you can still launch the Java server with our <a href="https://gradle.org/docs/current/userguide/gradle_wrapper.html" target="_blank">Gradle wrapper</a> script located at the root of your project:
+</p>
+<p>
+    <code>
+        gradlew bootRun
```
jdubois: I would do "./gradlew bootRun"

development.html
```diff
@@ -51,18 +51,30 @@
 <h3>(Optional) As a Gradle project</h3>
 
 <p>
-    If you selected the Gradle option when creating the project, you can launch the Java server with Gradle:
+    If you selected the Gradle option when creating the project, you can launch the Java server with Gradle if you have installed it:
```
jdubois: I would write it differently, as I would recommend using the wrapper instead of installing Gradle

gmarziou:
Why? I find it annoying to wait for first installation and also I don't have '.' in my $PATH which is I think a good security practice

gmarziou:
By the way I didn't write ./gradlew as I would have to write it also as .\gradlew for Windows

PR #46, opened 2015-05-11
Updated Heroku documentation to include heroku-deploy plugin

jkutner:
Hello, I work at Heroku and I'd love to help get this sub-generator out of beta. It's very clever in how it uses the "deploy/heroku" dir, but I think this can be avoided. There are three potential methods of deployment:

## 1. Git
This would involve adding the `Procfile` and `HerokuDatabaseConfiguration.java` from the generator, but it would need to get rid of the `.slugignore`. Then, the generator would set two buildpacks for the app:

```
$ heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs
$ heroku buildpacks:add https://github.com/heroku/heroku-buildpack-java
```

And add the following dependencies to the `package.json`:

```json
"dependencies": {
    "bower":"1.4.1",
    "grunt-cli":"0.1.13"
},
```

Finally, a `git push heroku` would push all source to Heroku, run the npm commands, and the maven build. This same approach could be used with the Gradle buildpack in place of the Maven buildpack.

## 2. CLI
This is what I've described in the PR. It primarily consists of this command:

```
heroku deploy:jar --jar target/*.war --options "--spring.profiles.active=prod --server.port=\$PORT --spring.datasource.heroku-url=\$DATABASE_URL"
```

## 3. Maven
The [heroku-maven-plugin](https://github.com/heroku/heroku-maven-plugin) can be used in lieu of the CLI plugin. It requires a new Maven dependency:

```xml
<plugin>
    <groupId>com.heroku.sdk</groupId>
    <artifactId>heroku-maven-plugin</artifactId>
    <version>0.4.0</version>
    <configuration>
        <includeTarget>false</includeTarget>
        <includes>
            <include>target/*.war</include>
        </includes>
        <processTypes>
            <web>java $JAVA_OPTS -jar target/*.war --spring.profiles.active=prod --server.port=$PORT --spring.datasource.heroku-url=$DATABASE_URL</web>
        </processTypes>
    </configuration>
</plugin>
```

And then the app can be deployed with a command like this:

```
mvn heroku:deploy
```

## Discussion

If these seem acceptable to you, I will make a PR against the generator to support, hopefully in the least, the CLI model. If the Git and Maven models are desired I will improve it to handle those as well. 

Thanks for the attention, and thanks for creating the sub-generator. There are number of folks using it!

Here is a sample project I've been using:
https://github.com/kissaten/jhipster-example


jdubois:
Hi @jkutner !
As you work for Heroku you probably know this a lot better than I do: please go ahead, and I will review your work later this week.
I would love to have something running smoothly on Heroku. By the way, could I have some free credits for JHipster, so I can have a couple of running apps on the "paying tier"? This would be only for testing, and showing people how it works, of course. I'm currently doing all my demos with CloudFoundry, but if I could do a "multi cloud" deployment during my demos that would be even better.

jkutner:
Cool, feel free to merge this then (it's still relevant). And I'll work on the generator PR. Thanks!

PR #47, opened 2015-05-15
Updated Heroku docs for new generator

jkutner:


jdubois:
Thanks, I will merge this when the next release is ready (so the website is on sync with the release)

PR #55, opened 2015-05-24
Add tip for QueryDSL

omrzljak:


deepu105:
@omrzljak the links in the tip doesnt work and also I guess you did a replace all for querydsl with Querydsl so in the code its capitalized and doesnt work

omrzljak:
1. If you check liks in https://github.com/jhipster/jhipster.github.io/blob/master/tips/003_tip_add_querydsl_support.md 
they are all correct and working. Could you be more specific on which link is not working good?

2. yes, you are wright , all Querydsl in build.gradle should be lowercase. But I don't know how it comes that it is capital case. If you look to my commit file https://github.com/jhipster/jhipster.github.io/pull/55/files you can see all are lowercase. But after merge of @jdubois  in master there are capital case ???

PR #62, opened 2015-06-15
Update configuring_ide_idea.md

Zyst:
Extending IDEA Settings as requested.


Zyst:
Do apologize for not placing the images in the /images folder. 

Thought it'd be better to just post those links directly and avoid messing with the meat of the project. Note that those links _will_ die after ~3 months, so if the screenshots are relevant they should be saved locally. or moved to a mirror that does not delete content.

Cheers!

PR #71, opened 2015-07-13
Create 006_tips_enable cross_origin_requests

tomcgn:
as requested from @deepu105

### DIFF
--- a/tips/006_tips_enable cross_origin_requests
+++ b/tips/006_tips_enable cross_origin_requests
@@ -1,0 +1,26 @@
+# Access to the API from 3rd party sites
+
+You might want to use the nice out-of-the-box API to integrate your app into existing websites. 
+One approach could be to use [Knockout](https://github.com/knockout/knockout) to bind your entities with ease and render the entity on e.g. your blog, being served from your jhipster app. 
+two aspects need to be changed in standard jhipster in order to use the API from simple HTML/AJAX clients:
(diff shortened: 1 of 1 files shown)


jdubois:
Thanks!!!! I'll review this next week.

gmarziou:
I think the part about the CORS headers filter is a bit hacky as it modifies an existing filter CsrfCookieGeneratorFilter which has a different purpose. It has 2 drawbacks:
- it violates single responsibility principle
- it makes updating jhipster alit bit more difficult

tomcgn:
Hello Gael,
Thanks for the feedback.
You're right. I had a suspect that it might be so. Unfortunately I have
not enough understanding of the exisiting Filters  and CSRF in general
to propose a better solution.

Do you have an idea how to achieve the same goal: accessing the jhipster
api from an independent, 3rd party website / JS client?
Then, apart from the valid design principles reasons you pointed out: do
you see security aspects undermined with the current approach?

Kind regards,

Thomas


Am 07.08.2015 um 17:51 schrieb Gaël Marziou:

gmarziou:
Don't worry Thomas, in fact I deleted my comment few hours after I wrote it because it was not so relevant and also when JHipster will support Spring Boot 1.3 we will rewrite this tip to use new annotation @CrossOrigin introduced in Spring 4.2.

PR #85, opened 2015-08-29
remove the maven recommendation

deepu105:
remove the default recommendation for maven

give maven and gradle equal chance

### DIFF
--- a/creating_an_app.html
+++ b/creating_an_app.html
@@ -45,6 +45,7 @@ mvn spring-boot:run
 <code>
 mvn
 </code>
+</p>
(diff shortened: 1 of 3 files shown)


creating_an_app.html
```diff
@@ -41,6 +41,21 @@
 mvn spring-boot:run
 </code>
 </p>
+<p>or simply
```
jdubois: I don't see where you close this tag

deepu105: Oops my bad, ill correct that

deepu105:
corrected :+1:

PR #90, opened 2015-09-17
Update maven command for new deployment

rodikal:
change the maven goal to 'package' for the following reasons:
* there is no need to install the package in the local maven repository, you just need a packaged war file
* this makes the page in sync with the "Using in production" page


jdubois:
Agreed!
Please, next time, follow our guidelines -> there should be a ticket first before doing a PR

PR #97, opened 2015-10-09
Plural Function not found for locale

gezero:
After following the guide to install new language I faced an error when loading the translations. I solved it by adding the proper local javascript file for my local into the index.html


jdubois:
This looks related to https://github.com/jhipster/generator-jhipster/pull/1916 from @dimitrisli

deepu105:
btw this should be closed here :)

PR #107, opened 2015-11-07
Added tip: Configuring HTML 5 mode

brevleq:
To configure this mode in JHipster, we need more than only set the html5mod flag, we need a Servlet Filter to redirect the urls without #, so activation link can work correctly.


jdubois:
Great, thanks!

PR #102, opened 2015-10-21
Installation guide changes for docker setup.

indyaah:
Fixes jhipster/generator-jhipster#2191

PR #109, opened 2015-11-09
Updated docker docs

sircelsiussexy:
Hi,

I add a few problems setting up docker today. After a while I found jhipster/jhipster-docker#16 that explained a few things.

I removed all mention of using SSH to access the container from the docs and added information on how to connect to the container with the latest Docker version.


sircelsiussexy:
Nevermind, it seems #102 already takes care of updating the docker docs.

PR #115, opened 2015-11-21
JDL relationship clarification

tuxetuxe:
I could only figure it out by reading the Jhipster-UML code :/

### DIFF
--- a/jhipster_uml.md
+++ b/jhipster_uml.md
@@ -568,9 +568,8 @@ The relationships declaration is done as follows:
 - `<relationship name>` is the name of the relationship in the entity.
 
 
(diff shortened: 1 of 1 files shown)


jdubois:
@MathieuAA do you confirm?

MathieuAA:
For me it's clear enough, but I'll have a more inquisitive look.

MathieuAA:
My bad, I thought there was a problem with the implementation.
The documentation doesn't explicitely (there's only one *hidden* example) say how it can be done with the UML editors, so I guess you can mege it, yes. I'll make a PR with the changes for the UML editors.

MathieuAA:
Update, and a big sorry. You can merge it but it is already explained in the documentation how to perform such a thing, in the very beginning of the doc right [here](https://jhipster.github.io/jhipster_uml.html#umlfile).

PR #117, opened 2015-11-21
Add tip to regenerate all entities at once

cbornet:
See https://github.com/jhipster/generator-jhipster/issues/2265

### DIFF
--- a/creating_an_entity.html
+++ b/creating_an_entity.html
@@ -149,6 +149,13 @@ lastmod: 2014-10-10T00:00:00-00:00
         <li>You have copy/pasted the .json file, and want a new entity that is very close to the copied entity</li>
     </ul>
 </p>
+<p>
+    TIP : to regenerate all your entities at once, you can use the following commands (remove the <code>--force</code> to have questions asked when files have changed).
+    <ul>
+        <li>Linux &amp; Mac: <code>for f in `ls .jhipster`; do yo jhipster:entity ${f%.*} --force ; done</code></li>
+        <li>Windows: <code>for %f in (.jhipster/*) do yo jhipster:entity %f --force</code></li>
+    </ul>
+</p>
 
 <h2>Tutorial</h2>
 


jdubois:
What a great tip!!! It is going to save me a lot of time! We should make this a command in the sub-generator

cbornet:
Yes we should ! What is your preference ?
```shell
yo jhipster:entity  (without args)
yo jhipster:entity regenerate 
yo jhipster:entity --regenerate
```
Or something else ?

jdubois:
Not `yo jhipster:entity regenerate` as this should create the "regenerate" entity.

The `yo jhipster:entity  (without args)` looks easy to use, but maybe we should have a question at the very beginning, warning of what is going to be done? Like "are you sure you want to regenerate all your entities?"

And also, it would be better to regenerate the entities using the timestamp that is stored in each of the JSON files, so they are regenerated in the same order -> I'm not sure this would case any issue as of today, but it looks risky to regenerate them in another order

cbornet:
```regenerate``` is not an entity since it begins with a lower-case letter but it can be confusing, I admit.
I guess yo jhipster:entity with the question is the best solution. We can even skip the question if ```--force``` is used.
I've always regenerated without taking care of the order and it has always worked fine. Since the regeneration doesn't modify the order of the files loaded by liquibase, I think it's not a problem and it's much simpler to implement.

**Comment by cbornet** [comment]
We could indicate the feature as ```[BETA]``` in the question

**Comment by deepu105** [comment]
what about yo jhipster:entity --all we can avoid the question then?

**Comment by deepu105** [comment]
or even better we can have new sub gen `yo jhipster:regenerate --all | entity` so that original entity subgen will not be disturbed at all

**Comment by cbornet** [comment]
Well, since I'm struggling to add it to the entity generator, I kinda like your idea of a new subgen. The only thing is that we are not really generating anything and using yeoman for this kind of things is a bit overkill... Maybe this could be a simple js script.
Also I think it should be ```--force``` is better than ```--all``` to be compliant with the other commands.

**Comment by deepu105** [comment]
ya correct, we can just trigger the commandline to run that one liner script

On Tue, Dec 1, 2015 at 12:00 AM, Christophe Bornet <notifications@github.com
> wrote:

> Well, since I'm struggling to add it to the entity generator, I kinda like
> your idea of a new subgen. The only thing is that we are not really
> generating anything and using yeoman for this kind of things is a bit
> overkill... Maybe this could be a simple js script.
> Also I think it should be --force is better than --all to be compliant
> with the other commands.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jhipster/jhipster.github.io/pull/117#issuecomment-160671045>
> .
>



-- 
thank you,
Deepu.k.s

**Comment by deepu105** [comment]
my idea was like this

`yo jhipster:regenerate --all`

will trigger `yo jhipster && for f in `ls .jhipster`; do yo jhipster:entity ${f%.*} ; done`

`yo jhipster:regenerate --entity`

will trigger `for f in `ls .jhipster`; do yo jhipster:entity ${f%.*} ; done`

if `yo jhipster:regenerate --all --force` then we just pass the --force to the command

**Comment by deepu105** [comment]
our entity subgen is very complex already so we should move towards the plugin kind of style what say @jdubois

**Comment by cbornet** [comment]
Oh, I see ! You mean regenerate the whole project ! This is an idea !
Then the ```--``` should be removed from ```all``` and ```entity``` as they are not options but rather arguments.

**Comment by cbornet** [comment]
Other possibility:
* Regenerate all
```
yo jhipster:regenerate
```
* Regenerate only entities
```
yo jhipster:regenerate --entities
```

deepu105:
ya that looks better as well, user can decide to pass --force if needed.
this could be our first plugin actually :)

PR #124, opened 2015-11-28
Added explanations to use devTools in IntelliJ

bourdux:
Since Eclipse and IntelliJ work differently and following jhipster/generator-jhipster#2242, I updated the documentation to show how to use devtools in IntelliJ IDEA.

### DIFF
--- a/configuring_ide_idea.md
+++ b/configuring_ide_idea.md
@@ -3,7 +3,7 @@ layout: default
 title: Configuring your IDE
 sitemap:
 priority: 0.7
-lastmod: 2015-05-22T18:40:00-00:00
+lastmod: 2015-11-28T17:13:00-00:00
 ---
 
 # <i class="fa fa-keyboard-o"></i> Configuring Intellij IDEA
--- a/configuring_ide_idea.md
+++ b/configuring_ide_idea.md
@@ -57,3 +57,9 @@ Navigate to `Languages & Frameworks → Javascript → Bower` (or type "Bower" o
 Point to your `bower.json`, which is located at the root of your project. The project's libraries, like Angular.js, should be automatically recognized.
 
 After configuring this you should have fairly extensive code support for the Javascript libraries in JHipster.
+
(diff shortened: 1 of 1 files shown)


jdubois:
That's a bit strange, I don't have this issue: with default settings, I just click on the "make" button, on the left of the Application class in the toolbar. It works without any issue.

Can you check this out?

Of course, this can be improved, for some time I was mapping one of my mouse's button to do this automatically (that was before Razer messed up with their driver support on my Mac - hopefully the new Logitech Proteus Core I should have for Christmas will solve that!)

deepu105:
@bourdux you could also mention remapping `ctrl + s` to save and compile the class so that behaviour is similar to eclipse you can just do a `ctrl + s`
@jdubois If im not wrong the make button does the same thing he is mentioning in the tip `ctrl + F9`  the problem is all files are autosaved so doing a ctrl+s just does save all, it doesnt seem to compile the changed classes

bourdux:
@deepu105 Yes, that was my point. I will modify the text according to your comments.

@jdubois Yes, it works for me. But the problem is what @deepu105 explained above. IntelliJ does not compile on save.

jdubois:
Yes, that button is for compiling. I think I have the default Intellij setup here:

- My files are automatically saved.
- When I hit the "compile" button, Intellij then puts the new classes in the target/ directory
- As soon as Spring Boot devtools sees those new classes, it reloads the application

This works very well for me

bourdux:
@jdubois So alright, I see what you mean. Now that you say it, it is not really a specific JHipster issue so it does not really have to be in the JHipster doc.

The thing is that I saw this question asked in jhipster/generator-jhipster#2242 and someone suggested to document it. Since I had to solve the same problem, I decided to document it. But anyway, I'll leave you the final choice of including this section in the public doc or not.

jdubois:
Yes I'm merging this, I'm just going to change a couple of things afterwards

PR #150, opened 2015-12-30
Managing relationships page - direction[n]al typo fix & bi-directional to bidirectional fix & uni-directional to unidirectional fix

mhavrlent:
There is a typo in `Managing relationships` page in word directionnal. Should be directional (one n). Also in words bi-directional and uni-directional there should be no dash. Bi-directional should be bidirectional and uni-directional should be unidirectional. See https://en.wiktionary.org/wiki/unidirectional and https://en.wiktionary.org/wiki/bidirectional for reference.


jdubois:
Good catch, thanks!

PR #159, opened 2016-01-04
add elasticsearch-reindexer

geraldhumphries:
This module will add a service to reindex all rows for each JHipster entity in Elasticsearch. This is useful if any changes have been made to the data outside of the application, like inserting rows using liquibase or migrating data from another DB. It works by reading all the json files from the `.jhipster` folder.

This is a solution I developed for my own project and decided to make it a module. There might be a better solution out there, but so far I haven't found one.

### DIFF
--- a/modules/marketplace/data/modules.json
+++ b/modules/marketplace/data/modules.json
@@ -129,5 +129,15 @@
         "description": "Stateless Basic authentication support for the API",
         "keywords": ["basic", "authentication", "api"],
         "verified": false
+    },
+    {
+        "name": "Elasticsearch Reindexer",
+        "npmPackageName": "generator-jhipster-elasticsearch-reindexer",
+        "url": "https://github.com/geraldhumphries/generator-jhipster-elasticsearch-reindexer",
+        "author": "Gerald Humphries",
+        "authorUrl": "https://geraldhumphries.com/",
+        "description": "Add a service to reindex all JHipster entities in Elasticsearch",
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
good job @geraldhumphries you're the first non core dev team to code a module!

geraldhumphries:
Thanks @pascalgrimaud! I used your module generator.

@jdubois we should add the module generator to the "create a module" docs. It was super useful.

jdubois:
Yes, congrats!!

- You need to do more documentation on your README so that people understand what this is all about
- Yes you are correct for the module generator, but I haven't used it myself: can you do a PR for this?

geraldhumphries:
Okay. I've updated the docs to better explain what it is.

@pascalgrimaud how do you want you/your project to be referred to? Right now I have "Pascal Grimaud has created a [module generator](https://github.com/pascalgrimaud/generator-jhipster-module) that will create a sample module to be used as a starting point."

pascalgrimaud:
Oh it will be nice! Give me time to review my code. I did it very fast during happy holidays => so I prefer doing this PR, as soon as it is definitively ready! Is it ok for you @geraldhumphries ?
But it seems to work fine because it works for you!

geraldhumphries:
Yeah, that's fine. I will open an issue/pull request on your project if I have any suggestions for how it can be improved. I will let you create the pull request here when you think it's ready.

PR #163, opened 2016-01-08
update docs for hooks

deepu105:
documentation update for hook mechanism
have also updated to show required Jhipster version in the modules list


jdubois:
@deepu105 I'm moving the "required version" part to the "details" page, the table is becoming too big...

deepu105:
Ok

PR #170, opened 2016-01-18
Jhipster redesign

deepu105:
A complete redesign of the Jhipster website

* A cool new theme
* Cool new home page
* Improved mobile views etc


deepu105:
@jdubois I could still see old files in the repository master, may be while
merging those were not removed. Ill do a new PR to remove those

PR #171, opened 2016-01-18
More wallpapers

atomfrede:
Some wallpapers using the new blue from the landing page

I've uploaded the [sources into a repository](https://github.com/atomfrede/jhipster-artwork)
@jdubois If you like we can transfer it to jhipster namespace


jdubois:
Thanks @atomfrede 

- yes it could be good to have the sources in the images directory (like I do for https://github.com/jhipster/jhipster.github.io/blob/master/images/wallpaper-001.xcf - this is a Gimp file)
- anyway I'm merging this, cool!

atomfrede:
Oh I didn't see the source file. Will put my sources there too (may into source directory) and do another pr.

PR #178, opened 2016-01-21
use npm search & update module details

deepu105:
update module details to show readme file from github repo
update to use npmsearch.com to discover modules
added a module config files to maintain blacklists and verified modules map

fix https://github.com/jhipster/generator-jhipster/issues/2665
fix https://github.com/jhipster/generator-jhipster/issues/2660


deepu105:
I will update this to use npmsearch instead of modules.json as well

deepu105:
this is ready

PR #191, opened 2016-02-01
Removed limitations warning for Heroku

jkutner:
Heroku now has a 90 second boot timeout limit for Java apps. I have found that this is sufficient for the majority of JHipster apps. /cc @mraible

### DIFF
--- a/pages/heroku.html
+++ b/pages/heroku.html
@@ -83,7 +83,9 @@ heroku deploy:jar --jar build/libs/*war
 <h2>Troubleshooting</h2>
 
 <p>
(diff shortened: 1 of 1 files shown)


jdubois:
Still, I'd like to leave the comment about the locked table, as this is really confusing to users when they don't know about this.
This could just be a "tip" or "comment" on that page.

jkutner:
@jdubois it's still mentioned in the "Troubleshooting" section. But I haven't seen that happen in a while. I think async liquibase stuff ensures that it happens well before any boot limit.

jkutner:
@jdubois I've updated the Troubleshooting section to include some of the old language. I also found that it need some tweaking.

PR #196, opened 2016-02-12
update blacklist

deepu105:
This module is a WIP so need to blacklist for now

### DIFF
--- a/modules/marketplace/data/modules-config.json
+++ b/modules/marketplace/data/modules-config.json
@@ -3,6 +3,6 @@
         "generator-jhipster-fortune" : true
     },
     "blacklistedModules": {
- 
+        "generator-jhipster-webservice" : true
     }
 }


cbornet:
@deepu105 what is your idea for this module ? Is it something like https://github.com/cbornet/generator-jhipster-swagger-cli. Maybe the two can be merged ?

deepu105:
Ok in my last project i wrote a custom jhipster sub generator to generate
soap webservice client stubs with jaxb, mapstruct and spring webservices.

I also did one to do rest webservice consumption from server side using
spring rest template. Thought of adapting that into a module as its a
common requirement in enterprises to integrate with soap and http services
from backend.

Im not sure how this fits in with swagger, coz im not very familiar with
swagger cli

cbornet:
Swagger-cli does approximately the same thing with the swagger.json that jaxb does with the WSDL.
I think the two WS description types (WSDL and swagger) could fit in the same module

deepu105:
You mean swagger can generate clients for wsdl as well?

cbornet:
No. Swagger is a specification for REST API  just like WSDL is a specification for SOAP API.
Swagge-codegen is a software that can generate code from a swagger spec but it can't do anything with WSDL.
And my module is just a wrapper around swagger-codegen to integrate this code in a JHipster app.
I was just saying that we could have a module that wraps both swagger-codegen and a "jaxb-codegen" depending on which spec the user has for the remote API.

deepu105:
Oh now I got it, let me have a look at your module. In that case i can skip
doing restClient part for http and use swagger codegen for that rite?

cbornet:
Probably. From what did you generate your REST client ? If it's from a WSDL, then maybe it still has an interest for people that do REST with WSDL (even if they really ought to switch to swagger for this...).
In that case I would suggest to harmonise and use [Netflix's Feign library](https://github.com/Netflix/feign) everywhere as it's a really awesome lib for back-end.
Also, I should probably move this module to hipster-labs. What do you think ?

cbornet:
In fact, for one of my projects I have an http API described with WSDL so I have personnal interest in your module :smile:

deepu105:
Yes you can move if you like so that there will be more people to maintain
it

deepu105:
I was generating a generic rest client skeleton for http services without
any schema actually
On 12 Feb 2016 19:37, d4udts@gmail.com wrote:

Yes you can move if you like so that there will be more people to maintain
it
On 12 Feb 2016 19:03, "Christophe Bornet" <notifications@github.com> wrote:

Probably. From what did you generate your REST client ? If it's from a
WSDL, then maybe it still has an interest for people that do REST with WSDL
(even if they really ought to switch to swagger for this...).
In that case I would suggest to harmonise and use Netflix's Feign library
<https://github.com/Netflix/feign> everywhere as it's a really awesome lib
for back-end.
Also, I should probably move this module to hipster-labs. What do you think
?

—
Reply to this email directly or view it on GitHub
<https://github.com/jhipster/jhipster.github.io/pull/196#issuecomment-183275886>
.

PR #208, opened 2016-03-06
Add minimum required version of Java

idNoRD:
Users need to know which Java version they can use with jhipster. It needs at least 8 version of Java. The 7 version is deprecated for jhipster.

### DIFF
--- a/pages/installation.html
+++ b/pages/installation.html
@@ -26,7 +26,7 @@ sitemap:
 
 <h2>Local installation (recommended for normal users)</h2>
 <ol>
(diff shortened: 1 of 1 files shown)


jdubois:
OK, but that's kind of obvious, no? JDK 7 has been deprecated by Oracle for several years now, it would be strange to install an old and unsupported version, and expect to have it working.

idNoRD:
Users still can use purchased [commercial long-term support](http://www.oracle.com/technetwork/java/javase/downloads/eol-135779.html).
If JHipster works only since Java 8 I think it is better to mention that, especially in the manual of setup environment.

jdubois:
I'm adding this in the v3 version of the documentation

PR #225, opened 2016-03-21
Add note about separate dev/prod assets

erikkemperman:
Added a note about separate dev/prod web assets, as requested in [#2785](https://github.com/jhipster/generator-jhipster/issues/2785) (checkbox 22).

I wasn't really sure what is expected here, so please let me know if I got this completely wrong, or incomplete, or whatever!

Looked at the "Using in development" section as well, but that already has a bit about this separation and I think that's good enough?

One thing I noticed: it says unconditionally in the production page that the WAR files will be generated as `target/jhipster-0.0.1-SNAPSHOT.war`, but I think that's only actually true for Maven?

### DIFF
--- a/pages/production.html
+++ b/pages/production.html
@@ -32,6 +32,10 @@ gradlew -Pprod
 This profile will compile, test and package your Java application with all productions settings. It will also trigger a "gulp build", so your static resources are optimized.
 </p>
 <p>
+Earlier versions of JHipster put these optimized assets in <code>src/main/webapp/dist</code>, as a child directory of the "dev" assets. As of 3.0, these have been separated and
+the optimized assets are now located in <code>target/www</code> for Maven or <code>build/www</code> for Gradle.
+</p>
+<p>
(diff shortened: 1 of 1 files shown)


pages/production.html
```diff
@@ -32,6 +32,10 @@
 This profile will compile, test and package your Java application with all productions settings. It will also trigger a "gulp build", so your static resources are optimized.
 </p>
 <p>
+Earlier versions of JHipster put these optimized assets in <code>src/main/webapp/dist</code>, as a child directory of the "dev" assets. As of 3.0, these have been separated and
```
jdubois: I will re-phrase this - we don't talk about "earlier versions" or we will end up with tons of older stuff

erikkemperman: Ah, yes, that seems sensible.

erikkemperman:
Note, the remark about prod/dev Frankenstein war-files should be removed if the pr for default spring profiles were merged.

PR #237, opened 2016-03-26
Updated the JDL page to include new things; plus some order changing

MathieuAA:
Updated the page to include all the latest things about the JDL

### DIFF
--- a/pages/jdl.md
+++ b/pages/jdl.md
@@ -4,7 +4,7 @@ title: JHipster Domain Language
 permalink: /jdl/
 sitemap:
     priority: 0.5
-    lastmod: 2016-03-13T12:00:00-00:00
+    lastmod: 2016-03-26T12:00:00-00:00
 ---
 
 # <i class="fa fa-star"></i> JHipster Domain Language (JDL)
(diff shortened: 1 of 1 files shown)


MathieuAA:
The output's not the same between the local version and the one displayed by GitHub. I'll have a PR ready to fix it ASAP. Sorry

pages/jdl.md
```diff
@@ -57,19 +70,33 @@ The entity declaration is done as follows:
...
+entity C {}
+entity D {
+  name String required,
+  address String required maxlength(100),
+  age Integer required min(18)
+}
+```
 
-Here's an example of a field declaration with validations:
+Because the JDL was made to be simple to use and read, f your entity is empty (no field), you can just declare an entity with `entity A` or `entity A {}`.
```
deepu105: @MathieuAA does this work in JHUML 1.6.1 ?

pages/jdl.md
```diff
@@ -77,39 +104,59 @@ The relationships declaration is done as follows:
...
+    entity Book
+    entity Author
+    
     relationship OneToMany {
       Author{book} to Book{writer(name)}
     }
 
-### Relationships
-The relationship OneToMany A to B is equivalent to the relationship ManyToOne B to A, you only need make one of them.
+Of course, in real cases, you'd have a lot of relationships and always writing the same three lines could be tedious.
```
deepu105: @MathieuAA this as well?

pages/jdl.md
MathieuAA: no, v1.6.2

pages/jdl.md
MathieuAA: the same, v1.6.2

PR #244, opened 2016-03-30
Adding Jekyll Docker command in README

wmarques:
I think we should add the @pascalgrimaud amazing docker command in readme, it's much easier to run Jekyll this way instead of installing it manually.


pascalgrimaud:
@wmarques : are you falling in love with Docker ? :whale: :smile:

wmarques:
@pascalgrimaud Haha yes I've to admit that it's beautiful to run a simple command and see it working at first time ! I wasted like 30min just to install the correct version of ruby to work with jekyll...
So yes : :whale: :whale: :whale: :+1: :+1: :+1:

PR #245, opened 2016-03-30
Remove CORS Tip

wmarques:
As asked in [this issue](https://github.com/jhipster/generator-jhipster/issues/3287) in generator-jhipster by @cbornet


jdubois:
Can you leave the page (but keep it removed from the index page), and write in big under the title that is has been deprecated?
This is to prevent broken links from Google, which would hurt our page rank.

wmarques:
@jdubois it's done

PR #248, opened 2016-04-01
add new tip. how to use mariadb

flanciskinho:
add new tip to use mariadb on jhipster

closed #3315


jdubois:
Seeing how easy it is, maybe this should be better as a new option in the generator?

flanciskinho:
give me a couple of days to try to do it.

PR #252, opened 2016-04-08
Clarifies the installation of NodeJS for Linux continuous configuration.

gmarziou:
After side discussion in https://github.com/jhipster/generator-jhipster/issues/3356


mraible:
:+1:

PR #253, opened 2016-04-13
Added MariaDB info and updated development info

adelimon:
Added MariaDB info and updated development info.  Added info about creating schema for a dev database, and username and password.

### DIFF
--- a/pages/development.html
+++ b/pages/development.html
@@ -191,7 +191,7 @@ sitemap:
     <img src="{{ site.url }}/images/h2.png"/>
 </p>
 
-<h3>Using MySQL or Postgresql in development</h3>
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Is it duplicated: https://github.com/jhipster/jhipster.github.io/pull/248 ?

adelimon:
It looks similar but isn't exactly the same.  #248 has more detail about MariaDB, and added to the tips documentation.  

I added my info to the main documentation, and also added a note about the default credentials, as well as creating the schema.  Minor things that popped up while I was working with this.  Those are small things that are easy enough to figure out though.

jdubois:
For me this is a duplicate.
Otherwise it doesn't add enough new info: as you say MariaDB is a fork of MySQL so our guide is true for both of them.

PR #264, opened 2016-05-13
To make jhipster console to run on Mac machine

sendilkumarn:
@PierreBesson Please note

### DIFF
--- a/pages/monitoring.md
+++ b/pages/monitoring.md
@@ -20,7 +20,7 @@ To configure a JHipster application to forward their logs to JHipster Console, e
         logging:
             logstash:
                 enabled: true
-                host: localhost
+                host: localhost #Replace this with docker's vm ip ( for Mac )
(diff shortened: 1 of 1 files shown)


jdubois:
Yes this is "normal" on Mac if you use Docker Machine, I'm not sure we need to explain that - and besides this will soon change with Docker for Mac and Docker for Windows, this works like on Linux on "localhost" and solves that issue.

PierreBesson:
@jdubois
It was me that asked for this to be PR to the docs.
I still think we should document it because people shouldn't need to be docker expert to evaluate the console.

@sendilkumarn can you replace  "(for Mac)" by "(if using docker-machine to create the docker host)". I think this issue also happen on windows. Then I will merge this. Thanks for the PR.

jdubois:
Merged but I changed some of the wording afterwards

PR #273, opened 2016-05-25
Jhipster upgrade

deepu105:


mraible:
Does this work for upgrading 2.x versions to 3.x?

cbornet:
This would be a miracle :smile:

mraible:
@cbornet I figured as much. The biggest issue I've had trying to upgrade from 2.x to 3.x is I was using xauth in 2.x and since we moved to jwt in 3.x, there's no equivalent and a lot fails. I've decided starting over from scratch is the easiest thing to do.

jdubois:
As the sub-generator is named "upgrade", I'm replacing "update" by "upgrade" in the title and text.

PR #275, opened 2016-05-26
Add a Ldap configuration tip

mleneveut:
Hi,

I've had trouble setting up the Ldap configuration as authentication, here is a proposal to add a Ldap section in the security page.


jdubois:
As this isn't working by default, I'd rather have this as a tip -> http://jhipster.github.io/tips/
Could you refactor this and create a new tip?

mleneveut:
Moved the security section in the tips page.

PR #276, opened 2016-05-31
Added new microservices and gateway video in pages/video_tutorial.html

baskeboler:


jdubois:
thanks @baskeboler but your video still doesn't have any voice comments (or notes), so it's very hard to understand what you are doing. I thought you added some?

baskeboler:
hi @jdubois , I just double-checked and the annotations are visible in the page.

jdubois:
Oh sorry @baskeboler !!

- I thought it was sound comments -> doing annotations might be better as many people watch those from work, so that's good
- However I didn't see them as I removed all comments on Youtube, as they are usually spam -> maybe you could add this in your description, something like "You need to have Youtube annotations ON to see the comments"

PR #277, opened 2016-06-07
Corrected docker invocation

lrkwz:
Tip won't work with wrong name case.

PS I've requested a pull in the main project in order to create a docker-file for this


pascalgrimaud:
thanks @lrkwz for the typo

PR #280, opened 2016-06-19
Update docker_compose.md

MathieuAA:
Updated the page to include a solution for a permission error when installing unbundled NPM packages.

### DIFF
--- a/pages/docker_compose.md
+++ b/pages/docker_compose.md
@@ -68,6 +68,8 @@ A solution would be to do:
   - `npm install -g YOUR_PACKAGE`,
   - then exit and go into container with normal way: docker exec -it jhipster bash.
 
+Further more, the NPM doc recommends not installing any NPM package as root. Follow the [link](https://docs.npmjs.com/getting-started/fixing-npm-permissions) to fix it.
+
 The original solution was proposed by Pascal Grimaud over [there](https://github.com/jhipster/jhipster-uml/issues/193).
 
 
--- a/pages/docker_compose.md
+++ b/pages/docker_compose.md
@@ -6,7 +6,7 @@ redirect_from:
   - /docker_compose.html
 sitemap:
     priority: 0.7
-    lastmod: 2016-03-20T00:00:00-00:00
+    lastmod: 2016-06-19T00:00:00-00:00
 ---
 
 # <i class="fa fa-music"></i> Docker and Docker Compose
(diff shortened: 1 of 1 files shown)


pages/docker_compose.md
```diff
@@ -61,6 +61,16 @@ If you are using Docker Machine on Mac or Windows, your Docker daemon has only l
 
 </div>
 
+
```
PierreBesson: I think we should put a title here like `Using JHipster UML in Docker`

pages/docker_compose.md
```diff
@@ -61,6 +61,16 @@ If you are using Docker Machine on Mac or Windows, your Docker daemon has only l
...
+If you encounter such an error `npm ERR! Error: EACCES: permission denied` when installing JHipster UML (or any unbundled package), your container may not have sudo (for instance, sudo isn't bundled with Ubuntu Xenial).
+A solution would be to do:
```
jdubois: I usually point people to https://docs.npmjs.com/getting-started/fixing-npm-permissions - isn't this better to go to the official NPM documentation?

pages/docker_compose.md
MathieuAA: If it does only happen with JUML, then yes, by all means.

Le 20 juin 2016 09:58:38 GMT+02:00, Pierre Besson <notifications@github.com> a écrit :
>> @@ -61,6 +61,16 @@ If you are using Docker Machine on Mac or Windows,
>your Docker daemon has only l
>>  
>>  </div>
>>  
>> +
>
>I think we should put a title here like `Using JHipster UML in Docker`
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/jhipster/jhipster.github.io/pull/280/files/1dc7c082bbe76bc42f82323d2390d80ff0516c8d#r67648853

-- 
Mathieu ABOU-AICHI

pages/docker_compose.md
MathieuAA: That's what I proposed to the original poster of the issue in JUML's issue tracker. I'll add it as a note to my side note.

pages/docker_compose.md
PierreBesson: I think it would also happen for people that want to install a JHipster module in docker.

MathieuAA: So, for people not having sudo installed and who want to install a package... I'd rather put a generic title, something like "Installing packages".

PR #282, opened 2016-06-22
Add extra information to exclude directories section

jmayday:
* added note about Intellij Frameworks and Libraries support


jdubois:
I need to check, but I would be even more explicit! I need to test, I'm not excluding those libs most of the time.

PR #283, opened 2016-07-01
Adding sections for memory tweaking and docker stats

danielbrutti:
Hi,

I just added information about a docker stats command and a new section for memory tweaking in case someone else had the [issue](https://github.com/jhipster/generator-jhipster/issues/3791) I had.

Thanks,

Daniel

### DIFF
--- a/pages/docker_compose.md
+++ b/pages/docker_compose.md
@@ -182,9 +182,9 @@ You can use `docker ps -a` to list all the containers
     fc35e1090021        mysql               "/entrypoint.sh mysql"   4 seconds ago       Up 4 seconds        0.0.0.0:3306->3306/tcp   sampleApplication-mysql
 
 ### Docker stats for containers
-`docker stats` or `docker stats $(docker ps --format={{.Names}})` to list all running containers with CPU, Memory, Networking I/O and Block I/O stats.
+`docker stats` or {% raw %}`docker stats $(docker ps --format={{.Names}})`{% endraw %} to list all running containers with CPU, Memory, Networking I/O and Block I/O stats.
 
-    $ docker stats $(docker ps --format={{.Names}})
+    $ docker stats {% raw %}$(docker ps --format={{.Names}}){% endraw %}
     CONTAINER                 CPU %               MEM USAGE / LIMIT     MEM %               NET I/O               BLOCK I/O             PIDS
     jhuaa-mysql               0.04%               221 MB / 7.966 GB     2.77%               66.69 kB / 36.78 kB   8.802 MB / 302.5 MB   37
     00compose_msmongo-app_1   0.09%               965.6 MB / 7.966 GB   12.12%              121.3 kB / 54.64 kB   89.84 MB / 14.88 MB   35
@@ -218,17 +218,13 @@ Be careful! All data will be deleted:
 
 ## <a name="8"></a> Memory Tweaking
 
-In order to optimize memory usage for applications running in the container, you can setup Java memory parameters on Dockerfile or docker-compose.yml
+In order to optimize memory usage for applications running in the container, you can setup Java memory parameters on `Dockerfile` or `docker-compose.yml`
 
 ### Adding memory parameters to Dockerfile
 
 Set the environment variable.
 
-    ENV JAVA_OPTS="-Xmx512m -Xmx256m"
-
-Or modify CMD line.
-
-    CMD ["java","-Xmx512m", "-Xmx256m", (...)]
+    ENV JAVA_OPTS=-Xmx512m -Xmx256m
 
 ### Adding memory parameters to docker-compose.yml
 
@@ -238,4 +234,10 @@ Add the JAVA_OPTS into `environment` section.
 
     environment:
       - (...)
-      - JAVA_OPTS="-Xmx512m -Xmx256m"  
+      - JAVA_OPTS=-Xmx512m -Xmx256m
+
+We found situations where `JAVA_OPTS` does not work and you need to use `_JAVA_OPTIONS` instead. In this case, the `environment` section looks like:
+
+    environment:
+      - (...)
+      - _JAVA_OPTIONS=-Xmx512m -Xmx256m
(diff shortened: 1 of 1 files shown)


danielbrutti:
Hi.. I found something.. setting just the environment section on docker-compose.yml doesn't work for me

- JAVA_OPTS="-Xmx512m -Xmx256m"

Please, do not approve this doc changes yet.. will continue working on this.

pascalgrimaud:
Probably a problem in syntax, with double quote

danielbrutti:
Is working, but I need to add `${JAVA_OPTS}` into Dockerfile CMD line on each project to make sure the opts are added to java -jar command.

I can add this into the docs.. but I'm wondering if there is another way to make sure it uses JAVA_OPTS without modify Dockerfile.

danielbrutti:
Also.. the problem arises when executing the registry, because the memory parameters are not used when registry is in docker compose:
~~~
jhipster-registry:
        extends:
            file: jhipster-registry.yml
            service: jhipster-registry
        environment:
            - JAVA_OPTS="-Xmx512m -Xmx256m"
~~~

and jhipster-registry.yml defines the image from docker hub.

~~~
image: jhipster/jhipster-registry:v2.2.0
~~~

In this case, I don't have control on CMD line to setup ${JAVA_OPTS}

danielbrutti:
OK found it.. it works when setting up this "_JAVA_OPTIONS" environment variable... this one is used by default by the JVM. If you agree, I can modify docs using this variable instead of JAVA_OPTS

PierreBesson:
@danielbruti It's strange, I thought _JAVA_OPTIONS was the old name of the variable. Also it shouldn't be necessary to modify the Dockerfile to edit the env var from docker-compose. But if it happens to be the case, I will add this to the jhipster's app and registry's Dockerfile on the CMD line.
 I will check everything tomorrow and come back to you.
Cc @pascalgrimmaud

danielbrutti:
@PierreBesson setting up `_JAVA_OPTIONS` in docker-compose.yml is not necessary to add the variable in Dockerfile. CMD can remain as it is right now.

PierreBesson:
Ok. I guess we can go with _JAVA_OPTIONS.

pages/docker_compose.md
```diff
@@ -199,3 +214,28 @@ When you stop a container, the data is not deleted, unless you delete the contai
 Be careful! All data will be deleted:
 
 `docker rm <container_id>`
+
+
+## <a name="8"></a> Memory Tweaking
+
+In order to optimize memory usage for applications running in the container, you can setup Java memory parameters on Dockerfile or docker-compose.yml
```
jdubois: Can you use backquotes around filenames? Like `docker-compose.yml`

pascalgrimaud:
strange! I always use JAVA_OPTS

pascalgrimaud:
@danielbrutti can you try exactly (without -> " ), and no additional "command" in docker-compose file
```
        environment:
            - JAVA_OPTS=-Xmx512m -Xmx256m
```

jdubois:
Is this ready to be merged or not? Doesn't look like you agree on the comments... And I also did one comment in the code that wasn't closed.

danielbrutti:
Hi, I'll finish this today, had hard weeks on work. 
Sorry for the delay.

danielbrutti:
Just pushed last changes. It did not work with 
~~~
- JAVA_OPTS=-Xmx512m -Xmx256m
~~~
So I left `_JAVA_OPTIONS` in docs too.

pages/docker_compose.md
danielbrutti: Done.. I also modified raw data. `docker stats {% raw %}$(docker ps --format={{.Names}}){% endraw %}`

jdubois:
I didn't know about this `_JAVA_OPTIONS` setting, and found some info on [this StackOverflow question](http://stackoverflow.com/questions/28327620/difference-between-java-options-java-tool-options-and-java-opts).
I'm not a big fan of it, it's an undocumented option, and `JAVA_OPTS` should work everywhere -> I would like to remove the part on this

danielbrutti:
OK, I will remove it.. I'm not a fan of use anything different of JAVA_OPTS... but for some reason it didn't work for me. However, I'm ok with remove _JAVA_OPTIONS since is not the standard.

pascalgrimaud:
@danielbrutti : for your information, with my current customer, we manage to run a full stack microservices with Docker on our test environment. I didn't put any specific configuration about memory, everything works perfectly !

danielbrutti:
@pascalgrimaud yes, I read is working for everyone, just failing to me :P 
That's why I'm ok with remove that option from docs.. because is more probably to be an issue from my side or OS or anything else, completely different to JHipster + Docker.

danielbrutti:
I just removed the options from docs.

pascalgrimaud:
@danielbrutti : your analyze helps me a lot on my current project, we had to optimize our memory on our machines. So big thanks!
You were totally right about JAVA_OPTS. It didn't work on docker-compose, but _JAVA_OPTIONS works!

I will analyze and do more tests, then I will PR an update on this documentation.

danielbrutti:
@pascalgrimaud : glad to hear that it did helped you!
We are still using _JAVA_OPTIONS, I really don't know why it does not work with JAVA_OPTS.. indeed we must use it as parameter when starting the registry with docker run too:
```
docker run -e "_JAVA_OPTIONS=-Xmx512m -Xmx256m" -p=8761:8761 -it jhipster/jhipster-registry:v2.2.0
```
Let me know if there is anything I can help.

pascalgrimaud:
I can see Xmx, 2 times. Maybe you miss Xms ?

danielbrutti:
@pascalgrimaud yes you are right, is Xms. Thanks!

jgasmi:
The JAVA_OPTS works fine for me as shown below in case this may help

`environment:
            - "JAVA_OPTS=-Xmx512m -Xmx256m"`

PR #288, opened 2016-07-13
Polish and improvements

deepu105:
A nice logo slider for companies using JHI in home screen
Optimized all the images and added `gulp images` task for the same (needs to be run manually when required)
polish companies using jhipster page to show nice tiles
other general improvements

image

@jdubois you can merge this right away if you are ok with the updates


jdubois:
Thanks! Lots of image downloads - this isn't very good but I don't have a good solution right now, and that's just static hosting by Github so it should be OK

deepu105:
The images are optimised anyways so shouldn't be lot of bandwidth. Ofcourse
if the companies using is gonna grow tremendously we need to try something
better

Thanks & regards,
Deepu

jdubois:
The problem is that we already have 115 resources loaded on the main page...

deepu105:
May be we can do some gulp tasks to concat and minify js and CSS but still
most of the number would be images and Im not sure if we can do much about
it. IMO loading individual images are much faster than loading a huge
sprite with all of them in it and in both cases browser will cache them
anyway.

Currently im not seeing any performance lag and that too im on a very bad
internet :P

Thanks & regards,
Deepu

PR #293, opened 2016-08-14
add fisherman plugin instructions

atomfrede:
@jdubois Not sure where to put the link (sidebar or just a link/hint in the oh-my-zsh page).


jdubois:
I'm not using fisherman so I'm not the best person to review this...

- We could move it to jhipster/jhipster-fisherman-plugin if you want to make it "official", and of course I'll give you full rights on the repo
- We could put it in the "setting up your environment" section, but I'm not sure a lot of people know about fisherman (I didn't!), so this might confuse people

atomfrede:
Moving it to jhipster would be cool. I also think adding it to the sidebar is not required as zsh is much more popular than fish shell.

jdubois:
@atomfrede I just created https://github.com/jhipster/jhipster-fisherman-plugin and gave you admin rights to the repo.

Indeed putting it in the sidebar would confuse people, but then where do we put this documentation?? Maybe along the zsh plugin?

atomfrede:
Cool. Will move the plugin to that repository. I would add it to zsh page or rename the sidebar entry to something like shell plugins

jdubois:
Yes OK for the sidebar entry "shell plugins" (note to self: make a redirection from oh-my-zsh to this new page)

atomfrede:
@jdubois I've update the links and added a parent page ``Shell Plugins`` to the sidebar which lists both plugins.

PR #300, opened 2016-08-22
Added Deploying to AWS with Boxfuse section

axelfontaine:
Please let me know if you want anything added or changed.


jdubois:
Thanks @axelfontaine - at first sight it looks OK, I'll read everything (for typos etc....) before merging, but that should be quick

jdubois:
I'll just rewrite the titles -> "AWS with Beanstalk" is just too long on our menu... So I'll keep "AWS" as before (as Beanstalk is the "normal" AWS), and rename your page "Deploying with Boxfuse"

axelfontaine:
Thank you Julien. The reason I added AWS with Beanstalk is that "normal" AWS can now mean CodeDeploy, CodePipeline, Beanstalk or OpsWorks! The number of different ways they offer to solve the same problem is really staggering and this just helped clarify the one being used here.

jdubois:
Yes I kept it on the main page, but on the left menu that's just too long

axelfontaine:
Would you be OK with a quick PR that cross references both?

Something like this (+links of course):

On the Boxfuse page: 

> As an alternative to Boxfuse you can also deploy your JHipster application to AWS using Beanstalk

On the Beanstalk page:

> As an alternative to Beanstalk you can also deploy your JHipster application to AWS using Boxfuse

jdubois:
Yes -> don't hesitate to advertise for Boxfuse! You provide full JHipster support, AWS doesn't!

PR #307, opened 2016-09-10
Proposed workflow for working with Mapstruct DTOs + Maven + IDE

amr:
More testing would be appreciated, while it works for me, I'm not sure this covers other cases / work-flows I might not be using / aware of.

### DIFF
--- a/pages/configuring_ide_eclipse.md
+++ b/pages/configuring_ide_eclipse.md
@@ -72,3 +72,11 @@ At this stage you should not have any Java error but should still see some JavaS
 ![Exclude]({{ site.url }}/images/configuring_ide_eclipse_6.png)
 
 ![Multiple select]({{ site.url }}/images/configuring_ide_eclipse_7.png)
+
+### Maven IDE profile
+
+If you are using Maven, you need to activate the `IDE` profile in IntelliJ. This is used for applying IDE-specific tweaks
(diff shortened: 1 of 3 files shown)


jdubois:
Thanks @amr !! That's a good idea:

- How about generating that profile, for everyone?
- I wonder how this works in Eclipse
- And of course we also need the same for Gradle

amr:
Those are pretty good points :) I could work on generating the profile for everyone as well as gradle support if I got a chance tomorrow but if not, then would probably work on that when I'm back from a 1 week trip.

Regarding Eclipse, the instructions are IDE agnostic, I just named it `intellij` as an example, the important part is activating the profile in the IDE and that's certainly possible with Eclipse.

cbornet:
The IDE configuration pages should also point to that tip (since new devs will read them before the mapstruct/dto page)

jdubois:
Oh yes, for the IDE you are correct, it's just me being tired!

amr:
PR updated with @jdubois suggestions. Profile implementation PR is jhipster/generator-jhipster#4118

The PR above as well as the updated docs herein do not include any special instructions for Gradle as I found that `mapstruct-processor` is being applied already with all Gradle builds (via `apply from: 'gradle/mapstruct.gradle'`), not profile dependent -- or maybe I missed something.

Can someone take a look and just confirm that this is the case?

PR #308, opened 2016-09-11
Fix mongodb logo

pascalgrimaud:
Trying to find a better logo
Currently, the mongodb logo is black on main page


jdubois:
Yes but your new logo doesn't have the correct size

pascalgrimaud:
yes, different size with the original, but in the img/icons/ folder, every images have a different size without problem when rendering on home page

If you prefer, I can try to rework this image to keep the same size as initial

PR #311, opened 2016-09-13
replaced gradelw with gradlew on creating-an-app page

mhavrlent:
There is a typo on https://jhipster.github.io/creating-an-app/ page: "gradelw". This PR replaces it with correct command: "gradlew".


jdubois:
Thanks!!

PR #312, opened 2016-09-14
fix the docker command to serve the jekyll website

PierreBesson:
Fix https://github.com/jhipster/generator-jhipster/issues/4131


pascalgrimaud:
I wonder if this looooong command should be add to our [zsh](https://github.com/jhipster/jhipster-oh-my-zsh-plugin) ? (if yes, we should test if a `Gemfile` is present before, to avoid wrong launching on another folder than jhipster.github.io)

PR #314, opened 2016-09-20
added better docs for mock testing uaa components

xetys:


pages/using-uaa.md
```diff
@@ -305,32 +305,40 @@ This approach addresses a scenario when machine request run over a separate OAut
...
+        //assert that your application is in the desired state
+    }
 }
 ```
 
-Since declaring that implementation as a usual Spring component, all Beans injecting a client will inject this implementation, so you can focus on the logic of these Beans.
-
-***Don't forget to declare these tests are running using profile "test" using e.g. `@ActiveProfile`***
+So with this technology you are simulating the behavior of the other service, and provide expected resource entity which would come from the origin. 
+All Beans winjecting a client will behave as mocked, so you can focus on the logic of these Beans.
```
jdubois: winjecting -> injecting?

PR #320, opened 2016-09-27
eclipse: update screen folder

pascalgrimaud:
Tested with latest Eclipse Neon.
The instruction are still correct. So I only updated the screenshot
Hope you are ok, my OS is ubuntu so the selected folder is now orange!

Fix https://github.com/jhipster/generator-jhipster/issues/4201


jdubois:
Argh we're doing the same thing at the same time :-(
I just commited this also!

jdubois:
OK mine is better, as I also changed the text, see https://github.com/jhipster/jhipster.github.io/commit/70971c2957d851f36ec25f2fd94d3f52fed7b6e5

pascalgrimaud:
ahah, I admit your screen is better :)
It was an easy PR just before going to bed

jdubois:
Yes, same for me :-)

PR #321, opened 2016-09-27
add note to querydsl tip to not include extra slf4j dependency

ruddell:
Fix jhipster/generator-jhipster#3908


jdubois:
Thanks @ruddell !!

PR #333, opened 2016-10-19
Remove json ext. when update entities on windows

Dufgui:


jdubois:
I can't test as I don't have Windows, so I'm trusting you :-)

PR #341, opened 2016-11-14
Tip: kubernetes google sql

bourdux:
I wrote a tip/tutorial to show how JHipster can be used with a managed Google Cloud SQL instance through Kubernetes. Since I fiddled a lot while writing this tutorial, external testing is very welcome :)


jdubois:
- Looks great!
- I'm merging this & tweet about it as it's the best way to have feedback
- Indeed, having a sub-generator to automate is the best solution, so this might be a first step. I am in contact with people at Google working on this, so I will follow this

PR #344, opened 2016-11-15
Add tip - Using docker containers as localhost on Mac/Windows

Akuka:


jdubois:
I'm using Docker 4 Mac (and I guess that's the same with Windows), and I don't have this issue anymore, did you try it?

pascalgrimaud:
I think it can be interesting for people who use Docker Toolbox (with Windows7).

PierreBesson:
Oh yes and remember that docker for windows only work with Windows 10 pro edition for now. So there are still many people stuck on the toolbox.

Also personally I like to use docker-machine to create virtual docker hosts (local or in the cloud) and clusterize them with swarm/swarm mode. On Mac, you can actually configure docker-machine to use xhyve which is the same as what docker for Mac use. On Linux you can do similar things with KVM.

pascalgrimaud:
Just tested this tips on Linux, with docker-machine and it works perfectly!

Akuka:
I'm using Docker Toolbox on my Mac and I have this issue. (I didn't knew it is even possible until I saw your video from Devoxx, which you started mysql server as docker container and then just connected to it like it is localhost)

pascalgrimaud:
To be precised, it's not an issue at all. It's a normal behavior because Docker Toolbox is inside a VM!
But your tips is a nice improvement!

Akuka:
Right, not an issue, but less convenient to work that way, that why its a Tip :)

BTW, in addition to the manual steps that i mentioned inside the tip, there is an option to configure it with command line by executing the following command:
`VBoxManage modifyvm <vmname> [--natpf<1-N> [<rulename>],tcp|udp,[<hostip>], <hostport>,[<guestip>],<guestport>]`

For example, for MySQL:
`VBoxManage modifyvm default --natpf1 "mysql,tcp,,3306,,3306"`

jdubois:
OK, if you all like it, let's add it!

PR #347, opened 2016-11-18
Kubernetes: kafka support

pascalgrimaud:
- See https://github.com/jhipster/generator-jhipster/pull/4415
- Thanks to @raphaelbrugier


jdubois:
I will merge this when we do the next release, is this good for you?

pascalgrimaud:
It's already supported since the 3.11.0 !
So it can be merged now

PR #348, opened 2016-11-19
Update microservice architecture diagram & refactor duplicate image folder

deepu105:
@mraible you can find a high res image in `images/artwork` folder


jdubois:
The diagram is much too big, it doesn't fit on my screen, I'm reducing its size

PR #355, opened 2016-11-26
added jhipster ide

colameo:


colameo:
please feel free to adapt and change content

jdubois:
Thanks: I'll change some of the content, I hope that will be fine with you

And I'll add this to the menu on the left, of course!

jdubois:
I'm keeping the "eclipse2.gif" file but it is not used, shouldn't it be removed?

jdubois:
And for this to work, I need to remove your "gh" branch -> I'm doing it, I hope it's OK for you

jdubois:
Oh, I just needed to disable Github pages, so no worries!

colameo:
@jdubois the eclipse2.gif is actually the optimized version from @PierreBesson, so maybe it's better to use that one. I've spoken to @jbadeau this morning and he will make a nicer page with more content.
BTW, @cbornet will also add an installation guide (or a screen cast) about the visual studio code extension. Thanks for publishing it!

jdubois:
OK, can you replace the image then? And add content, modify, etc, like you feel like -> then it's just a PR and I'll merge this very quickly

PR #359, opened 2016-12-01
Added notes about the following comments

MathieuAA:
An example:
```
entity A {
  name String /** My super field */
  count Integer /** My other super field */
}
```


MathieuAA:
Thanks :) This time the doc is correct. I can't believe I didn't even know we could omit commas in the JDL. Must have been Carl who did it waaaaaay before the JDL was.. the JDL.

PR #368, opened 2017-01-03
How to run docker commands in the container

jmkhael:
Added instruction to how to run docker commands from within the container, for instance to be able to build docker targets with mvnw or gradlew.

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -169,6 +169,33 @@ If you copy-pasted the above command to run the container, notice that you have 
 `docker exec -it jhipster bash`
 
 You will login as "jhipster" user. In case you need to do a `sudo`, the password for the user is the same as the username (`jhipster`).
+An alternative for distribution which ships without sudo by default is to login with the user-id 0 like so:
+
+`docker exec -it --user=0 jhipster bash`
+
+### Docker commands in JHipster's docker container
+
+If you need to launch docker commands from within JHipster's container, e.g. to build the Dockerfiles `./gradlew bootRepackage -Pprod buildDocker`, you can use the `--group-add` flag to add the container user to the host's docker group id:
+
+    docker run --name jhipster \
+        --group-add 979 \
+        -v $(which docker):/usr/bin/docker \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ~/jhipster:/home/jhipster/app \
+        -v ~/.m2:/home/jhipster/.m2 \
+        -p 8080:8080 -p 3000:3000 -p 3001:3001 -d -t jhipster/jhipster
+
+Note that the docker group id (979 here) can change from distribution to another. Run `cat /etc/group | grep docker` to see yours.
+
+To start the container with the correct group in a single commands, you can execute:
+
+    docker run --name jhipster \
+        --group-add $(cat /etc/group | grep docker | awk -F":" '{print $3}') \
+        -v $(which docker):/usr/bin/docker \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ~/jhipster:/home/jhipster/app \
+        -v ~/.m2:/home/jhipster/.m2 \
+        -p 8080:8080 -p 3000:3000 -p 3001:3001 -d -t jhipster/jhipster
 
 ### Your first project
 


pages/installation.md
```diff
@@ -169,6 +169,33 @@ If you copy-pasted the above command to run the container, notice that you have
...
+
+If you need to launch docker commands from within JHipster's container, e.g. to build the Dockerfiles `./gradlew bootRepackage -Pprod buildDocker`, you can use the `--group-add` flag to add the container user to the host's docker group id:
+
+    docker run --name jhipster \
+        --group-add 979 \
+        -v $(which docker):/usr/bin/docker \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ~/jhipster:/home/jhipster/app \
+        -v ~/.m2:/home/jhipster/.m2 \
+        -p 8080:8080 -p 3000:3000 -p 3001:3001 -d -t jhipster/jhipster
```
pascalgrimaud: it should be 9000:9000 instead of 3000:3000
=> I know the documentation is not up-to-date, and I already changed it for JHipster 4

pages/installation.md
```diff
@@ -169,6 +169,33 @@ If you copy-pasted the above command to run the container, notice that you have
...
+
+To start the container with the correct group in a single commands, you can execute:
+
+    docker run --name jhipster \
+        --group-add $(cat /etc/group | grep docker | awk -F":" '{print $3}') \
+        -v $(which docker):/usr/bin/docker \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ~/jhipster:/home/jhipster/app \
+        -v ~/.m2:/home/jhipster/.m2 \
+        -p 8080:8080 -p 3000:3000 -p 3001:3001 -d -t jhipster/jhipster
```
pascalgrimaud: same: it should be 9000:9000 instead of 3000:3000

pages/installation.md
jdubois: Thanks @pascalgrimaud , I'm also updating this elsewhere in the documentation

jdubois:
Oh I see them now you submitted the review :-)
Then I don't understand anything anymore -> so we do have "sudo" installed, then what is this all about??

jmkhael:
Given that:

- Canonical decided that sudo is no longer part of their "minimal set of packages" ubuntu >= xenial. 
- the Dockerfile doesn't install it by default

sudo commands do not work out of the box. That's why I added the "+An alternative for distribution which ships without sudo by default is to login with the user-id 0 like so:"

Hope this clears it a bit

cbornet:
Then it's our previous instruction on using sudo that is wrong.
So we should either install sudo in the docker image or make ` --user root` the default command for running with permissions

jmkhael:
If we want to abide by the [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#user), it is best to go with option 2 when we need to run docker commands,

> If a service can run without privileges, use USER to change to a non-root user. 
> You should avoid installing or using sudo since it has unpredictable TTY and signal-forwarding behavior that can cause more problems than it solves. If you absolutely need functionality similar to sudo (e.g., initializing the daemon as root but running it as non-root), you may be able to use “gosu”.

jdubois:
I've been using 'sudo' for more than 20 years and I've never found it had "unpredictable" behavior...

jdubois:
I'm sorry but I'm lost now :-)
@cbornet could you correct the current text?

cbornet:
@jdubois the question is : do we install sudo in the docker container or do we document to login as root for privileged operations ? Both are fine for me also 😄 . 
I would even be fine with removing the whole sudo/root thing. I'm not sure why you would need it...

jdubois:
I find it incredible that "sudo" isn't part of the minimal packages anymore... But we can't do anything about this anyway... I still remember when you could login as root on Redhat, nearly 20 years ago, that was so much simpler!!!
As we don't have sudo installed yet, let's keep it that way (and we follow the "best practice"!), and document how to login as root. And if that's fine with you @cbornet then that's really perfect. Could you correct the current text? You seem to know this much better than I do.

cbornet:
OK, will do.

jdubois:
Thanks!

jdubois:
I've had a try at this, as this wasn't clear for me -> so it's quite easy, just use `--user root` to run as root, and then it's all the same as using `sudo`.

And then the part about running "Docker in Docker" doesn't work at all for me:

- I don't have a "docker" group in my Ubuntu image
- You can't run the command as "root" as Bower will blow up -> this will be solved with JHipster 4, when we remove Bower

So I'm rewriting the part about the `--user root`, and removing the part about "Docker in Docker" as it doesn't work. I'll link my commit here.

PierreBesson:
@jdubois I'm quite sure that docker in docker does work : it's just that ` -v $(which docker):/usr/bin/docker` to mount the docker binary in the container doesn't work with docker4Mac default config. Have a look at the docker4mac config screen were you can allow volumes to be mounted from /usr/bin/docker with the UI.
As for Bower, you can simply run  `bower install --allow-root`.

cbornet:
Yes. Docker in docker works fine. The part with root user has nothing to do with that.

jmkhael:
Hello, 
Have we lost the docker in docker part on the documentation for good now?

jdubois:
@jmkhael I removed it because it wasn't working: it isn't technically lost (still in Git), but we can't have a documentation that doesn't work.
This will change very soon anyway: for me the blocker issue is running Bower as Root, and this will disappear as soon as we switch to Angular 2.

jmkhael:
@jdubois I am lost, because the Docker in docker parts works quite fine without need for root as confirmed by @cbornet and @PierreBesson. As stated by @PierreBesson, the fact that it doesn't work seems related to docker4Mac settings to allow mounting /usr/bin/docker on your machine. This doesn't interfere with bower in any way.
On Linux (RHEL, Ubuntu and Raspbian) this works without any additional configuration.

pascalgrimaud:
Another option would be to do a tip, as the main installation page is
dedicated to all OS, and not something specific to Linux.

jdubois:
@jmkhael yes you can run Docker as root (I've done it without any issue), the problem is that Bower, with its default configuration, will not run as root (for security reasons). So of course we can force Bower to run as root with a specific option, and we don't really care about the security reasons anyway as we're in a container, but that's not what we currently do. And as Bower will disappear (with Angular 2 we switch to NPM/Yarn), I don't think it's worth it (and that will make our documentation a lot more complex in the end).

cbornet:
* As for login as root, I don't see why someone would want to do it. And that's standard docker usage so I don't think it's useful to add documentation about it.
* As for running docker from a docker container, that's not something easy to do so it would have value to document how to do it correctly. From what @PierreBesson says, binding the docker binary doesn't work on Mac (it works on Linux) so a generic and better solution would be to install the docker-cli inside the jhipster docker image. I can PR for that.

PierreBesson:
@cbornet, yes this was my initial idea to install the docker cli, but we must be sure to install the latest version. Also we should write in our docs that the user might have to set the DOCKER_API_VERSION if they don't have the same docker version on their host and in the container...

Although I'm not using the docker image for development myself I think it's important to support running docker commands from inside docker. So @cbornet please go on with the PR.

PR #375, opened 2017-01-18
Add tip Build profile dev/prod use different log file path

hookszhang:
If you want to enable the logging output to a file, there will be such a problem:

How do i according to the different environment config log file path?


pascalgrimaud:
It looks like quite complex -> the `logback-spring.xml` inside the generated project is provided as example, and you can override it.

For that, you should use the normal behavior provided with logback:
- look the properties about logging : https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html
- you can have different `logback-dev.xml`, `logback-prod.xml` etc
- when starting your app, you can do something like that:
    - `java -jar app.war --logging.config=/path/logback-prod.xml`

jdubois:
Yes, I'm sorry @hookszhang but I'd rather use a Spring Boot property, than doing a Maven/Gradle replacement. It's easier and it works in my IDE, without having to run a build tool. It's also easier to re-configure in prod.

hookszhang:
Thanks @pascalgrimaud @jdubois, you let me know some better ways.

jdubois:
Really sorry about this, I know you spent a lot of time to write this :-(

hookszhang:
That's all right. I am very happy, learning in a better way to solve this problem

PR #377, opened 2017-01-18
Updated Openshift.md

noelo:
Openshift has moved to the Kubernetes platform.

### DIFF
--- a/pages/openshift.md
+++ b/pages/openshift.md
@@ -11,6 +11,4 @@ sitemap:
 
 # <i class="fa fa-cloud-upload"></i> Openshift
 
-Openshift support has been removed, as Red Hat changed its APIs and nobody in the community was interested in maintaining this sub-generator.
-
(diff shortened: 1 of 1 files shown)


PierreBesson:
Well I'm a bit reserved on this. Although the Kubernetes subgen is indeed a way to deploy applications on Openshift v3. Yet I think this page should contain some information on how to deploy JHipster apps to Openshift with the Source-to-image method and other Openshift specific instructions.

Help would be appreciated if you have some experience deploying JHipster to Openshift as nobody on our team uses it.

jdubois:
I'm guessing there is some difference between Kubernetes and Openshift, so for me we don't support it at the moment.

jdubois:
If anybody wants to provide more information, and is ready to test it, as @PierreBesson asks, I'll be very happy to merge this. But at the moment I don't think we can say our Kubernetes support works with Openshift, nobody has ever tested this, as far as I know.

PR #380, opened 2017-01-27
Added doc about JDL's constants

MathieuAA:
As of JHipster Core v1.2.7, released just before this PR.


jdubois:
I will merge this when we release JHipster 4, if that's OK with you

MathieuAA:
No prob.

PR #381, opened 2017-01-30
Prepare jhipster 4

deepu105:
I'll adding some more docs so dont merge yet


jdubois:
That looks like a lot of work ! I'll try reviewing it today on the train.
I'll prepare the release notes.

jdubois:
@deepu105 I will replace the "NG1" and "NG2" references, as they should be "AngularJS 1" and "Angular" (maybe "Angular 2.x").
Don't update this PR as I will merge it locally, if you want to change more stuff, please do other PRs.

deepu105:
Ok, but I often find people getting confused when we say just AngularJS and
Angular as it not obvious hence we should be more clear and say AngularJS
v1 and Angular v2+ or something like that


Thanks & Regards,
Deepu

jdubois:
I don't understand why this wasn't closed automatically, for me this is merged -> that's why I'm closing this

PR #396, opened 2017-02-21
complement ci-cd docs with missing question

kaidohallik:


pascalgrimaud:
Maybe I was too tired!

PR #407, opened 2017-03-05
Added section about yarn, yarn.lock and yarn cache. Requested by http…

Zacorich:
…s://github.com/jhipster/generator-jhipster/issues/5362#issuecomment-284175620

### DIFF
--- a/pages/development.md
+++ b/pages/development.md
@@ -165,6 +165,32 @@ This will launch:
 - A BrowserSync task that will run on [http://localhost:9000/](http://localhost:9000/), which has a proxy to [http://localhost:9060/](http://localhost:9060/) (the Webpack "hot module reload" server), and which will synchronize the user's clicks/scrolls/inputs
 - The BrowserSync UI, which will be available on [http://localhost:3001/](http://localhost:3001/)
 
+### About yarn, yarn.lock and yarn cache(important):
+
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Following the docs https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-force

Did you try `yarn install --force` ?
> This refetches all packages, even ones that were previously installed.

Zacorich:
For my particular case solution was to regenerate yarn.lock file so I didn't try to re-fetch packages after I solved the issue. I just added more info about yarn to jhipster website for people to pay attention to yarn internal work and be aware of what can go wrong and where to look.

PierreBesson:
@Zacorich Is it possible to rework the phrasing a little. The point you make here is interesting but your sentences are too long, they should be broken down in small chunks.

jdubois:
@PierreBesson is right, I'm not merging this as is, for 2 reasons:

- There are some formatting issues, like missing upper cases (`Yarn` not `yarn`), etc... And indeed the sentences are too long
- We usually don't document third-party tools: here you give hints on how to use Yarn, and this isn't specific to JHipster

Then, I understand this would help many people, but we need to rework this. I'll have a look when I have time.

jdubois:
I have heavily modified your text, and I'm merging this, thanks!

PR #446, opened 2017-06-12
Proxy settings for Maven Wrapper

sonata82:
The Maven Wrapper script also requires proxy settings (as mentioned [here](https://github.com/rimerosolutions/maven-wrapper-example#proxy-server-configuration)) which are not pulled from the Maven settings file. 

I've opted for configuring this using a file, to keep the instructions system independent.


jdubois:
I didn't know this, thanks a lot!!

PR #451, opened 2017-06-26
config-server encryption and JCE information

abshkd:
as promised from [jhipster-registry/#149](https://github.com/jhipster/jhipster-registry/issues/149)


PierreBesson:
👍 @abshkd, have a look at this: https://github.com/jhipster/jhipster-registry/pull/157

PR #452, opened 2017-06-29
Fix link to marketplace

madmas:
Currently, the link in the "Creating an application" docs brings up a 404. 
This PR fixes it :-)


jdubois:
Good catch!!!! Thanks a lot

PR #450, opened 2017-06-24
Update module page

pascalgrimaud:
Following the discussion at https://groups.google.com/forum/#!topic/jhipster-dev/DogpMHZzf3w

The steps:
- review this page
- as soon as it's ok, I will do a new release of [generator-jhipster-module](https://github.com/jhipster/generator-jhipster-module)
- then, this PR can be merged

PR #455, opened 2017-07-06
Update 2017-07-06-jhipster-release-4.6.0.md

deepu105:
@jdubois I guess its nice to show the deprecation warning in release notes.
This should be merged after #450


deepu105:
#450 is merged

PR #462, opened 2017-07-18
adds more k8s docs, removes BETA label

xetys:


deepu105:
Nevermind I see that its taken care of

PR #469, opened 2017-08-01
Update customizing-bootstrap-4.md

deepu105:
for next release

### DIFF
--- a/pages/customizing-bootstrap-4.md
+++ b/pages/customizing-bootstrap-4.md
@@ -27,4 +27,8 @@ For example to use the border-radius mixin:
 ```
 Make sure you import only partials and not main sass files, otherwise you will end up generating duplicate css which might cause issues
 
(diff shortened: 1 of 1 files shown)


pages/customizing-bootstrap-4.md
```diff
@@ -27,4 +27,8 @@ For example to use the border-radius mixin:
 ```
 Make sure you import only partials and not main sass files, otherwise you will end up generating duplicate css which might cause issues
 
+To change the default bootstrap settings like colors, border-radius, etc, add or change the value of the property in the partial file `src/main/webapp/content/scss/_bootstrap-variable.scss`
```
mraible: These sentences don't appear to have punctuation at the end of them. Should they?

Also, this file doesn't exist. I just finished re-creating 21-Points Health with 4.6.2 and there's only a `content/scss/vendor.scss` file, not a `content/scss/_bootstrap-variable.scss` file. See https://github.com/mraible/21-points/tree/master/src/main/webapp/content/scss.

deepu105: I probably missed punctuation. This file was added to master so wont be available in older releases

PR #477, opened 2017-08-29
Added a new tip for using Kafka

eosimosu:


jdubois:
Thanks a lot!
I'm not merging this yet as I want to review it, but this is definitely lacking - maybe this should be part of the "official" documentation, and not just a tip.

eosimosu:
@jdubois Happy to contribute! 
You can decide where it might fit best although I haven't seen similar "documentation" with a complete example for other supported technologies like elasticsearch etc.

jdubois:
Thanks again @eosimosu !
Indeed we lack a lot of documentation :-)
I would be happy to have this as part as the "official" documentation at http://www.jhipster.tech/using-kafka/ - and I guess we could remove the warning message, as well as the limitation (as you do provide an example, even if it's not generated)
What do you think?

eosimosu:
@jdubois sure, let's have it there then :) hopefully i can contribute more.
Btw, I'm using Jhipster + Kafka in a monolithic production app and haven't faced any limitations.

jdubois:
Thanks!!! It is now live on http://www.jhipster.tech/using-kafka/

eosimosu:
@jdubois  great!

PR #486, opened 2017-09-23
Insert googleoff: index into all release notes

gmarziou:
Fix #jhipster/generator-jhipster#6375

### DIFF
--- a/_posts/2013-11-30-jhipster-release-0.1.0.md
+++ b/_posts/2013-11-30-jhipster-release-0.1.0.md
@@ -12,6 +12,8 @@ We have also moved the JHipster github repository to the newly created JHipster 
 
 [https://github.com/jhipster](https://github.com/jhipster)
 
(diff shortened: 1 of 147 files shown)


gmarziou:
I must add that release notes take 320 KB versus 440 KB for documentation pages.
With this PR, I estimate to have cut the release notes weight to half.
It still means that past information represents 1/3 of our corpus and so biasing search index.

Once we stop maintaining angular 1, oldest release notes will be obsolete and we should consider removing them from the index which would be easy due to timestamp in filenames.

jdubois:
Sorry @gmarziou but that doesn't work, it messes up the page formatting, I need to rollback!

gmarziou:
Sorry I should have tested with Jekyll, I could not imagine that an HTML comment would be interpreted : https://github.com/vmg/redcarpet/issues/268

PR #487, opened 2017-09-24
Update production .war location

BarbaricCorgi:


pages/production.md
```diff
@@ -37,8 +37,8 @@ Or when using Gradle:
 
 This will generate two files (if your application is called "jhipster"):
 
-*   `target/jhipster-0.0.1-SNAPSHOT.war`
```
pascalgrimaud: we should keep both lines: target and build/libs, as it concerns Maven and Gradle too

BarbaricCorgi: hmmm you're right indeed

BarbaricCorgi:
I just fixed the paths to include both maven and gradle styles

jdubois:
Great, thanks!

PR #490, opened 2017-10-01
Adding Beginning Spring Boot 2 book reference.

sivaprasadreddy:
I have a dedicated chapter on JHipster in my new Beginning Spring Boot 2 book. Adding a reference to the book might help developers to get started with JHipster using Spring Boot 2.


jdubois:
I will need to add a specific section on JHipster books, cc @deepu105 -> it should be well referenced, and have one page per book, where each book author can write what he wants/needs

I'll do this in a couple of months, when @deepu105's book is ready

deepu105:
That should be good


Thanks & Regards,
Deepu

PR #497, opened 2017-10-18
master

yingw:


yingw:
sorry, mis commit

deepu105:
@yingw I didn't know you maintained a Chinese version of the JHipster doc site at https://jhipster-china.github.io/ :)
@jdubois what do you think?

jdubois:
There's also https://jhipster.ru/ in Russian... As I have no idea who is doing those sites (maybe until now for China?) I'm not doing any advertisment for them, I'm afraid they could contain malware. I don't understand the languages, I have no idea what is being done on those websites. That's definitely strange.

PR #506, opened 2017-11-26
Opencollective

deepu105:
Add open collective info and a contribute section

To be merged along with https://github.com/jhipster/generator-jhipster/pull/6719


jdubois:
- Thanks @deepu105 this is a huge work!!
- We can't merge the sponsors part yet, as it's empty - it wouldn't look very good! I would put it higher in the page, and wait until we have a couple of sponsors
- I'm going to create a new menu entry on the left, where:
  - I will add all sponsor information, including what they gain by becoming a sponsor, at the beginning:
      - for all baker/sponsor: a thank you Tweet from @java_hipster
      - for all baker/sponsor sponsors: being on the front page
      - for gold sponsors they would also have their logo at the bottom of each page
   - I will also move the "thank you" page here, and provide links to the "contributing" and "code of conduct" documents, as it seems a lot of people don't find them easily

-> is that fine with you?

deepu105:
Yes thats good for me. We can keep the empty ones commented out for now

PR #513, opened 2017-12-05
Add a note on API first implementation classes

cbornet:
See https://github.com/jhipster/generator-jhipster/issues/6495


jdubois:
Awesome :-)

PR #529, opened 2018-01-12
fix menu Text, should be "News"

yingw:
fix menu Text, should be "News"


deepu105:
@yingw no its intentional

deepu105:
Thanks anyway

PR #532, opened 2018-01-22
Feature/aws-containers

ggotti:
PR adds in documentation for the the new aws-containers sub-generator, awaiting PR [here](https://github.com/jhipster/generator-jhipster/pull/7035). 

It re-arranges the AWS page to describe both generator example.


ggotti:
Based on feedback, I've made the following changes
- Added detailed description on delete behaviour.
- Changed networking diagram to include ALB instead of NLB.
- Called out issues with cache synchronisation.

PR #537, opened 2018-01-31
update security config to use kebab case and add missing filter-order param

ruddell:
The app will not compile with camelCase config.

```
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'security.oauth2.client.access-token-uri' in value "${security.oauth2.client.access-token-uri}"
```

### DIFF
--- a/pages/security.md
+++ b/pages/security.md
@@ -93,16 +93,17 @@ security:
         enabled: false
     oauth2:
         client:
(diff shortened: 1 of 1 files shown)


mraible:
LGTM!

jdubois:
Awesome, thanks!

PR #539, opened 2018-02-09
Remove obsolete tip about enforcing https

gmarziou:
This tip is from 2015 and applies to Tomcat, it also points to a blog that no longer exists. I found some useful answers on stackoverflow for spring boot and Undertow that depend on context, so I think it's safe to remove it.

I also changed the numbers of tips list to "1." so that future deletions do not require re-numbering.


jdubois:
Thanks for cleaning up!

gmarziou:
I have a problem with Jekyll, the tips page redirects forever. Either using it directly or in docker. I can see it works on the site but does this page works for you locally?

jdubois:
@gmarziou yes it works fine for me.
It's installed locally on my Mac with Homebrew + the Ruby gems that are required. It's quite awful as when I upgrade Homebrew it regularly crashes because of uncompatible Ruby stuff... But at the moment it's fine :-)

gmarziou:
Damn, I got same issue on Ubuntu and Windows, well I'll leave with it :)

deepu105:
Works fine for me in Fedora, I installed Ruby and other required deps
directly


Thanks & Regards,
Deepu

pascalgrimaud:
@gmarziou : did you test the new docker-compose ? I worked on it and it should work... I'm on Ubuntu 17.10

gmarziou:
I'm also on Ubuntu 17.10 and yes I tried the new docker-compose. In fact, the site works except few things like the tips page and the fontawesome icons. I'll have a look when I can.

pascalgrimaud:
do you use http://0.0.0.0:4000 ? Not localhost

gmarziou:
I don't remember I'll check

PR #546, opened 2018-03-03
[v5] fix Gradle build, using bootWar

pascalgrimaud:
See https://github.com/jhipster/generator-jhipster/commit/edcf0529bd444f9c4d03fc002bcb78642efaf731

I suppose it's for v5 but not sure.


deepu105:
its definitely v5

PR #557, opened 2018-04-08
fix deprecated config key

ruddell:
When launching the site locally with `docker-compose`, it outputs a deprecation warning:

       Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -4,7 +4,7 @@ exclude: [node_modules, bower_components, vendor]
 url: https://www.jhipster.tech
 kramdown:
   input: GFM
-gems:
+plugins:
   - jekyll-redirect-from
 whitelist:
   - jekyll-redirect-from


jdubois:
Thanks!

ruddell:
Yes please revert if it breaks things, we can just ignore the warning.  

Not sure why it doesn't work as Github [uses](https://pages.github.com/versions/) Jekyll v3.7.3 and this was changed in Jekyll v3.5.0 ([release notes](https://github.com/jekyll/jekyll/releases/tag/v3.5.0)).  

The related PR and comment confirming it was deprecated: https://github.com/jekyll/jekyll/pull/5130#issuecomment-235732112

jdubois:
Oh sorry, I'm running it locally, and must have an older version.

jdubois:
I just upgraded, no difference.... but Jekyll starts 3 times faster, so that nice 👍

ruddell:
When testing the redirects locally I see the issue @gmarziou mentioned https://github.com/jhipster/jhipster.github.io/pull/539#issuecomment-364594570 when clicking on either Tips'n Tricks or the Release Notes.   This PR doesn't affect that for me.

I wouldn't spend too much time here if it was working with the old config.

jdubois:
Yes, let's keep the current config and not worry too much, we have lots of other stuff to do at the moment

PR #559, opened 2018-04-13
Fix typo in href to Bootstrap 4 theming guidelines

ahpearce:
The link to the bootstrap documentation on theming is broken. It should be 'theming' rather than 'themings'.


jdubois:
Thanks a lot!

PR #560, opened 2018-04-13
Require Node 64-bit versionBecause we use `--max_old_space_size=4096`…

gmarziou:
…, our build fails with node 32-bit.


jdubois:
Thanks!

PR #573, opened 2018-06-09
Update docker_compose.md

kanuni89:
fix gradle docker build


jdubois:
I'm pretty sure this is `bootWar` with the new Gradle plugin for Spring Boot 2 - if you can point to the documentation or give more information, I'll be happy to have a look and merge, but for the moment I think this is correct

PR #574, opened 2018-06-09
Update profiles.md

kanuni89:
fix gradle task package production build

### DIFF
--- a/pages/profiles.md
+++ b/pages/profiles.md
@@ -18,7 +18,7 @@ JHipster comes with two [Spring profiles](http://docs.spring.io/spring-boot/docs
 
 Those profiles come in two different configurations:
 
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
It depends on the version of JHipster you used. 

In 4.x, it is indeed `bootRepackage`
In 5.x, it is `bootWar`, after this change https://github.com/jhipster/generator-jhipster/commit/edcf0529bd444f9c4d03fc002bcb78642efaf731

jdubois:
Yes the documentation page is on the "current" version, so it is correct (considering our BETA release is the current, which indeed isn't totally clear).
This is not really good today, I thought the v5 release would be stable much sooner. But this is probably one of the only changes, there's nothing big.

jdubois:
Oh and thanks a lot @kanuni89 for this, those kind of PRs are very important - we don't have enough time to check everything, user feedback is very very important

PR #585, opened 2018-06-22
Fix a typo on the Profiles page

RPrimrose:
Tiny fix, Profiles page incorrectly claims there are three additional profiles but there are only two now that "shell" is deprecated


jdubois:
Thanks @RPrimrose !

PR #586, opened 2018-06-24
update blueprint docs

deepu105:


modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Variables from configuration:
+
+You can access to configuration in `.yo-rc.json` which will consist of both the JHipster config and your blueprint config.
+
+### Global variables:
+
+You can use constants in [generator-constants](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js):
+
+```
```
mraible: Add language

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+    const javaDir = `${jhipsterConstants.SERVER_MAIN_SRC_DIR + this.packageFolder}/`;
+    const resourceDir = jhipsterConstants.SERVER_MAIN_RES_DIR;
+    const webappDir = jhipsterConstants.CLIENT_MAIN_SRC_DIR;
+```
+
+### Functions:
+
+You can use all functions in [generator-base](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js):
+
+```
```
mraible: Add language

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+    this.angularAppName = this.getAngularAppName(); // get the Angular application name.
+    this.printJHipsterLogo(); // to print the JHipster logo
+```
+
+**Note**: The functions in `generator-base.js` and variables in `generator-constants.js` are part of public API and hence will follow semver versioning. But other files like `generator-base-private.js`, `utils.js` etc will not follow semver versioning and might break method signature across minor versions.
+
+## Registering a blueprint to the JHipster marketplace
+
+To have your blueprint available in [the JHipster marketplace]({{ site.url }}/modules/marketplace/), you need to make sure you have the 2 keyword `yeoman-generator` and `jhipster-blueprint` in your published npm `package.json`.
```
mraible: 2 keyword > two keywords

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+jhipster --blueprint <blueprint name>
+```
+
+## Example
+
+The [JHipster Kotlin](https://github.com/jhipster/jhipster-kotlin) blueprint replaces most of the server side Java code with equivalent Kotlin code.
+
+It is our official blueprint that showcases how you can create your own blueprint.
+
+The [JHipster Sample Blueprint](https://github.com/hipster-labs/generator-jhipster-sample-blueprint) shows how a client sub generator can be overriden.
```
mraible: sub generator > sub-generator

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+
+Or, you can use the [JHipster blueprint generator](https://github.com/jhipster/generator-jhipster-blueprint) to help you to initialize your blueprint. Choose the sub-generators that you would like to override while answering the questions.
```
mraible: It would be helpful to show the command to use this below so users don’t have to leave the page.

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
+---
+layout: default
+title: Creating a blueprint
```
mraible: Capitalize blueprint since it’s a title

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+        get initializing() {
+            // Here we are not overriding this phase and hence its being handled by JHipster
+            return super._initializing();
+        }
+
+        // other phases of the sub generator
+    }
+```
+
+Any method beginning with _ can be reused from the superclass that is being extended, for example `ClientGenerator` in above sample.
```
mraible: in above example > in the example above.

modules/creating_a_blueprint.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+**Note**: The functions in `generator-base.js` and variables in `generator-constants.js` are part of public API and hence will follow semver versioning. But other files like `generator-base-private.js`, `utils.js` etc will not follow semver versioning and might break method signature across minor versions.
+
+## Registering a blueprint to the JHipster marketplace
+
+To have your blueprint available in [the JHipster marketplace]({{ site.url }}/modules/marketplace/), you need to make sure you have the 2 keyword `yeoman-generator` and `jhipster-blueprint` in your published npm `package.json`.
+If you find any entry in the marketplace which is not a JHipster module or blueprint, you can help to blacklist it by adding it to the `blacklistedModules` section of the [modules-config.json file](https://github.com/jhipster/jhipster.github.io/blob/master/modules/marketplace/data/modules-config.json) by doing a Pull Request to the [jhipster/jhipster.github.io project](https://github.com/jhipster/jhipster.github.io).
+
+Your blueprint will become "verified" if the JHipster team verifies it.
```
mraible: Would “official” be better here?

deepu105:
@mraible thanks, i'll update as per your comments, asap

modules/creating_a_blueprint.md
deepu105: if its third party we cant make t official, we thought of having a verification system for modules but we never did that for any module. Maybe its better to just remove this line

PR #590, opened 2018-06-29
change from prefer to mandatory

ganeshkrishnan1:
nodejs 10 does not work with latest jhipster; only the LTS version works.
Removed the word "prefer"


jdubois:
In that case we can write it's mandatory... But most of the time the non LTS version works, so I'm fine with "prefer". If it fails, then people need to fall back to the LTS.

PR #596, opened 2018-07-06
Update jdl.md

Tonterias:
Added filter example and link to show at least once how to do filtering and where to put it.

### DIFF
--- a/pages/jdl.md
+++ b/pages/jdl.md
@@ -280,7 +280,7 @@ Latest version introduces exclusions (which is quite a powerful option when sett
     paginate C with pagination
 
 
-With JHipster, you can also tell whether you don't want any client code, or server code. Even if you want to add a suffix to Angular-related files, you can do that in JHipster.
+With JHipster, you can also tell whether you don't want any client code, or server code. Even if you want to add a suffix to Angular-related files, you can do that in JHipster. [Filtering](https://www.jhipster.tech/entities-filtering/) options can be activated on a per entity basis: filter `<entity name>` or for all entities: filter `*`.
 In your JDL file, simply add these lines to do the same:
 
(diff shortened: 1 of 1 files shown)


pages/jdl.md
```diff
@@ -280,7 +280,7 @@ Latest version introduces exclusions (which is quite a powerful option when sett
     paginate C with pagination
 
 
-With JHipster, you can also tell whether you don't want any client code, or server code. Even if you want to add a suffix to Angular-related files, you can do that in JHipster.
+With JHipster, you can also tell whether you don't want any client code, or server code. Even if you want to add a suffix to Angular-related files, you can do that in JHipster. [Filtering](https://www.jhipster.tech/entities-filtering/) options can be activated on a per entity basis: filter `<entity name>` or for all entities: filter `*`.
```
ruddell: Small nitpick but I would put the code tags around the full JDL, so:

`filter *` instead of "filter `*`" 
`filter <entity name>` instead of "filter `<entity name>`"

Tonterias: Ok, i did the changes you suggest and commit. It that it or should I do something else? Thanks Jon.

pages/jdl.md
```diff
@@ -291,6 +291,7 @@ entity C
 skipClient for A
 skipServer for B
 angularSuffix * with mySuperEntities
+filter C
```
jdubois: I didn't know about this one, @MathieuAA can you confirm?

MathieuAA: Sorry didn't see anything as I came back from vacations and didn't have the time to use my computer since then... Yes it's correct.

jdubois:
Thanks @MathieuAA

PR #603, opened 2018-07-24
Remove note about H2 disk-based dev db being beta

RothAndrew:
H2 disk-based dev database is no longer in Beta, and works on Windows. Updated the documentation to reflect this.


jdubois:
Thanks a lot! I have no idea why we still have this old documentation on such an important page!

PR #608, opened 2018-08-14
fix docs on metrics reporters

PierreBesson:
Fix https://github.com/jhipster/generator-jhipster/issues/8052

### DIFF
--- a/pages/monitoring.md
+++ b/pages/monitoring.md
@@ -15,7 +15,7 @@ JHipster provides several options to monitor your applications at runtime.
 1. [Generated dashboards](#generated-dashboards)
 2. [JHipster Registry](#jhipster-registry)
 3. [JHipster Console](#jhipster-console)
-4. [Forwarding metrics to a supported third party monitoring system (JMX, Graphite, Prometheus)](#metrics-exporters)
+4. [Forwarding metrics to a supported third party monitoring system](#metrics-exporters)
 5. [Zipkin](#zipkin)
 6. [Alerting with Elastalert](#elastalert)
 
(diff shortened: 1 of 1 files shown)


pages/monitoring.md
```diff
@@ -165,25 +165,17 @@ You can then put this data in a JSON file in one of the `jhipster-console/dashbo
 
 If you have created useful dashboards and visualizations for your JHipster applications please consider contributing those back to the community by submitting a Pull Request on the [JHipster Console's GitHub project](https://github.com/jhipster/jhipster-console).
 
-## <a name="configuring-metrics-forwarding"></a> Forwarding metrics to a supported third party monitoring system (JMX, Graphite, Prometheus)
```
jdubois: I'm very surprised, Graphite and Prometheus are just not working anymore?? What a huge step backwards!!

cub-bsi: You can still configure Graphite export in `application.yml` (for micrometer) by setting keys in the `management.metrics.export` namespace. This will push the normal Spring metrics at least – if you’re using Dropwizard for your own metrics, it will probably not push those. (Even though for some reason Dropwizard metrics are named as the "underlying instrumentation library", whatever that means, in the doc on Graphite export.)
See:
http://micrometer.io/docs/registry/graphite
Prometheus export also seems to be generally supported, but there are no config keys documented, only Java code:
http://micrometer.io/docs/registry/prometheus

PierreBesson: As long as we have not moved our metrics to micrometer properly I would not advertise it.

PR #615, opened 2018-08-27
Documentation for Deploying JHipster on Azure

brunoborges:


jdubois:
Thanks @danielpetisme - I'll correct the `./mvnw` part, and for Gradle I'm really not the best person so I'll leave it that way for the moment (PR welcome, of course).

I'll add this to the main menu, so it's consistent with the other cloud deployment options, that looks good to me that way.

PR #627, opened 2018-10-01
Fix typo `security.user.password` vs `spring.security.user.password`

danielpetisme:
A minor typo.


jdubois:
Thanks @danielpetisme !

PR #637, opened 2018-10-08
Fix excludes for Visual Studio Code

rainerh:
- add build to search.exclude
- add .gradle to files.exclude


jdubois:
Thanks @rainerh !

PR #643, opened 2018-10-15
add note about gatling 2 compatibility

atomfrede:
The current gatling documentation pushes gatling 3. I could not find a download link to the current 2.x release on their homepage anymore, so we should make clear we currently support only 2.x and provide a download link via maven central. 

See https://github.com/jhipster/generator-jhipster/issues/8552#issuecomment-429716324


jdubois:
Thanks @atomfrede !! That's pretty bad that they don't provide an easy way to install Gatling 2!

pascalgrimaud:
Maybe we should try to use Gatling 3. But it's another subject and as always, we need to find time to do this

PR #648, opened 2018-10-19
Update with my electron jhipster module generator

amanganiello90:


jdubois:
I'm sorry but we haven't used this feature at all yet, I don't think this does anything at the moment

amanganiello90:
Hi,
I haven't problem to insert my jhipster module in the jhipster marketplace. The module is compliant with the keywords, but seems not to be recognized.  Here my package: https://www.npmjs.com/package/generator-jhipster-electron

deepu105:
This won't solve your issue. We need to check whats wrong here. Can you debug the network calls on the websites marketplace to see if your module is returned when querying to npms.io

PR #650, opened 2018-10-20
rename fisherman to fisher

atomfrede:
* replace all occurences of `fisherman` with `fisher`
* update links

updates https://github.com/jhipster/jhipster-fisher-plugin/issues/17

### DIFF
--- a/pages/fisher.md
+++ b/pages/fisher.md
@@ -1,0 +1,34 @@
+---
+layout: default
+title: Using Fisher
+permalink: /fisher/
+redirect_from:
+  - /fisher.html
+sitemap:
+    priority: 0.7
+    lastmod: 2018-10-20T18:40:00-00:00
+---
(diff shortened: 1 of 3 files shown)


jdubois:
@atomfrede you also need to add the redirect from the old pages so we don't get 404 (important for Google Webmaster Tools)

atomfrede:
I did not remove the old page, so it is still there. How do I add a redirect from `fischerman` to `fisher`? Just adding an additional `redirect_from` in the new page? (sorry no jekyll expert)

jdubois:
Yes - otherwise I'll do it, but not until 2 weeks at least

jdubois:
I'm merging and then I'll correct the redirect

jdubois:
Oh you already fixed this, thanks a lot @atomfrede !

PR #660, opened 2018-11-09
Fix angular cli config filename

gmarziou:
Since angular 6, the angular cli config file has been renamed from `.angular-cli.json` to `angular.json`

Ref jhipster/generator-jhipster#5558


jdubois:
Thanks @gmarziou !

PR #668, opened 2018-11-21
use new artwork

atomfrede:
* favicon is the bowtie
* The hipster is randomly selected
* The logo is now the bowtie
* The artwork page just points to our artwork repository
* Removed past events from event listing

The equation at `What is JHipster` is not yet perfect imho. Wanted to use the bowtie, but our current blue background is basically incompatible with the bowtie. You can only see the left part and I also think its against the logo guideline. So I use the family member head according to the hipster at the top. But they are quite different from height so looks not always perfect.

screenshot_2018-11-21 generate a full stack java angular react application in a few steps
screenshot_2018-11-21 generate a full stack java angular react application in a few steps 1
screenshot_2018-11-21 generate a full stack java angular react application in a few steps 2
screenshot_2018-11-21 generate a full stack java angular react application in a few steps 3


jdubois:
Thanks a lot @atomfrede !

PR #679, opened 2019-01-14
Add @MapsId explanation and example in docs

pmverma:
I have added explanation and example in docs where it seems suitable.

Please have a look and let me know if it can be improved.


jdubois:
Thanks a lot!!

murdos:
Since this feature is not released, I guess docs should not have been merged...

PR #682, opened 2019-01-14
add documentation on jib workaround for offline use

PierreBesson:


pascalgrimaud:
thanks @PierreBesson !
very usefull

PR #690, opened 2019-01-30
Fix typo and wording regarding data replication

thokuest:
Elasticsearch defines itself as document oriented database, thus the sentence "This is normal, as Elasticsearch is not a real database" has been removed. Instead, it is clarified that there's no automatic replication mechanism.

Relevant link: [Elasticsearch as a NoSQL Database](https://www.elastic.co/blog/found-elasticsearch-as-nosql)


jdubois:
I would argue that Elasticsearch is not a database: of course it has no transaction and no relationships (as many NoSQL databases do), but it can also easily lose data. So it's a really, really poor choice of a database, and that's the main reason we don't provide Elasticsearch as a primary database.

Anyway, I think that's not the main point of your feedback: thanks a lot for submitting this, I'm merging it!

PR #691, opened 2019-01-30
improve random number

albert-ziegler:
I've noticed `Math.random()` used in a sensitive context. However, its values are not safe against prediction, so if I understand the context correctly, it's preferable to use `Window.crypto` instead.

Full disclosure: I noticed this using [LGTM.com](https://lgtm.com/projects/g/jhipster/jhipster.github.io/alerts/?mode=list&tag=security), a code analysis platform which I also work for.

### DIFF
--- a/presentation/plugin/multiplex/index.js
+++ b/presentation/plugin/multiplex/index.js
@@ -36,7 +36,8 @@ app.get("/", function(req, res) {
 
 app.get("/token", function(req,res) {
 	var ts = new Date().getTime();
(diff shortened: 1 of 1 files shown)


jdubois:
Thanks a lot, but this is from reveal.js and this is not our code initially. I'm not confident in changing code from them without a good reason, and it's just our presentation slides, so nothing important on the security part, I'd rather keep it this way.

albert-ziegler:
Cheers Julien, that makes sense!

PR #697, opened 2019-02-04
Add NYC workshop

Falydoor:


pascalgrimaud:
@Falydoor : I think you'll need a new release of the blueprint Vue.js before this event, right ?
Don't hesitate to ask us.

cc @sahbi-ktifa @atomfrede @deepu105

Falydoor:
Yes that would be cool to have a new release few days before it!

PR #698, opened 2019-02-07
Remove link to deprecated tips about user with additional information

pascalgrimaud:
Close https://github.com/jhipster/generator-jhipster/issues/8648


jdubois:
Oh I didn't know this was outdated.

BTW we should make `User` a specific object in the JDL, so we can add fields and relationships to it.

pascalgrimaud:
@nagarajasr : you can look this ticket https://github.com/jhipster/generator-jhipster/issues/8648

As already said, tips are written by users.
We can't guarantee it will work fine. If it needed to be updated, don't hesitate to PR and update the tips.

Here, no one want to update it, that's why I removed the link

PR #677, opened 2019-01-08
Added tip 26: upgrade jhipster_upgrade branch

tibistibi:
I use these commands to upgrade the jhipster upgrade branch, i think it is worth sharing.

PR #703, opened 2019-02-18
Upgrade with newest code

tibistibi:
With this you can upgrade the jhipster_upgrade branche to use your latest code

I added a tip here but maybe more in view is better:

https://github.com/jhipster/jhipster.github.io/pull/677


jdubois:
@tibistibi is this the same as #704 ?

tibistibi:
yes but the other has an extra line at the bottom. this one should be removed.

sorry about that

jdubois:
No problem, thanks @tibistibi !

PR #704, opened 2019-02-18
upgrade multiple times with your latest changes taken into account

tibistibi:
With this you can upgrade the jhipster_upgrade branche to use your latest code

I added a tip here but maybe more in view is better:

https://github.com/jhipster/jhipster.github.io/pull/677


jdubois:
Thanks @tibistibi !

tibistibi:
guess what i'm doing today...


(upgrade jhipster and search for the git commands....)

PR #708, opened 2019-03-01
fix(production): change npm script to build prod

clakech:
was "webpack:build" instead of "build" so the generated asset was a 20Mb JS ^^


jdubois:
Indeed, `webpack:build` does a dev build, when `build` does a prod build, this isn't very user-friendly!!!

jdubois:
Thanks @clakech !

PR #709, opened 2019-03-04
Remove rancher docs

PierreBesson:
Linked to https://github.com/jhipster/generator-jhipster/pull/9368


jdubois:
We shouldn't remove a page like that as it would do 404 errors for the Google Crawler, let me change this

jdubois:
Oh this sucks because I also don't like to have a "I'm sorry it's removed" page just for Google... Let's keep it like you did.

PR #711, opened 2019-03-10
add support for java.time.Duration

massimosiani:
Update the docs for adding support to java.time.Duration.


DanielFran:
See https://github.com/jhipster/generator-jhipster/issues/9095

PR #712, opened 2019-03-14
Added a section to run as a service

edinhojorge:
Although there is the possibility to run directly, it starts in foreground. It normally runs in auto mode, but added a way to explicit generate the package with service mode and pointing to spring docs.

### DIFF
--- a/pages/production.md
+++ b/pages/production.md
@@ -102,7 +102,7 @@ Secure your application with:
 `chown jhuser:jhuser jhipster-0.0.1-SNAPSHOT.war
 sudo chattr +i your-app.jar`
 
-being jhuser the non root os account that will run the application, and you will be able to run like that:
+Considering `jhuser` a non-root OS account that will run the application, then the application can be run this way:
 
 `service jhipster start|stop|restart`
 
--- a/pages/production.md
+++ b/pages/production.md
@@ -97,11 +97,16 @@ Next, setup your init.d with:
 
 `ln -s jhipster-0.0.1-SNAPSHOT.war /etc/init.d/jhipster`
 
-and you will be able to run 
+Secure your application with:
+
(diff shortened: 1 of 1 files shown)


jdubois:
Thanks @edinhojorge !
I totally forgot to write this, as this is how I run start.jhipster.tech and I find it's an awesome way to run the app.
Give me some time to check my configuration and maybe add some more tips on this, before merging.

edinhojorge:
Great, take your time... I just finished my first jhipster app release and deployed into prod, and found almost nothing about that. It is a great tool and I believe I will contribute with some code and doc, and also some tutorials.
Thank you @jdubois

PierreBesson:
Although I agree that we need to have this in our docs. However I think we should recommend using a SystemD Unit rather than a SysVInit service as it is the modern way to do it almost all Linux distributions.

See https://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html#deployment-systemd-service

But I see no problem with having the 2 methods in our docs.

PierreBesson:
Also your page is missing the very important security instructions from: https://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html#deployment-initd-service-securing

edinhojorge:
I was trying to keep it simple, with the fastest way and pointing to the docs, but you are right as we are under production section, security is a must... Just added to the docs, please review. I had a hard time setting it right with SystemD and auto restart. Will work on that again and add it to this section next time.
Thanks @PierreBesson

PierreBesson:
Thank you @edinhojorge. Please take your Time for the systemD docs. I think you will need to remove the `embeddedScript` part from the pom.

jdubois:
Thanks a lot @edinhojorge !

PR #713, opened 2019-03-15
Add missing return

smalltimecharlie:
Add missing return


jdubois:
Thanks @smalltimecharlie !

PR #716, opened 2019-03-27
Algolia search feature

sahbi-ktifa:
Replace Google Search by Algolia DocSearch.


jdubois:
I don't understand why, but the search box isn't perfectly aligned

PR #718, opened 2019-03-28
Update OIDC docs for JHipster 6

mraible:
Please don't merge yet, I'm still polishing...


mraible:
This is now ready to merge.

mraible:
@jdubois Can you please merge this? I don’t have permission.

jdubois:
Oh yes of course!
BTW we also need to release a new JHipster Registry version with this new configuration

mraible:
@jdubois I upgraded JHipster Registry in https://github.com/jhipster/jhipster-registry/pull/334.

jdubois:
Yes but I didn't release a new version, as it doesn't work with the docker-compose sub-generator.
I did update to the new configuration, but it fails with a 404 error on /login
I need to fill a proper ticket for that.

mraible:
That sounds like the client needs to be updated to point to /authorize. Sorry I can’t fix; I only have my phone right now.

jdubois:
Yes, same for me! It's not urgent, it works with the older release

PR #719, opened 2019-03-29
Move search bar css to global instead of home

sahbi-ktifa:
This is important as currently other pages than Home have their search bar broken.


jdubois:
I'm trusting you on this one, I'll let you test in prod :-D

jdubois:
Looks still broken on sub-pages :-(

PR #722, opened 2019-04-03
Add algolia doc search on mobile

sahbi-ktifa:
Available on top of the home page.


jdubois:
Awesome! @sahbi-ktifa this should be live pretty quickly, can you test? If you still see the old version in a few minutes, ask me and I'll force to clear the Cloudflare cache

PR #727, opened 2019-04-25
Improve link to the chinese website

PierreBesson:
Screenshot 2019-04-25 at 14 15 24

Screenshot 2019-04-25 at 14 15 35

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -105,6 +105,9 @@
                             <a href="{{ site.url }}/sponsors/">Sponsor</a>
                         </li>
                         {% endif %}
+                        <li>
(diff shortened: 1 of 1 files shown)


jdubois:
I'd rather keep the language at the bottom:

- That's how most websites do today, like https://docs.microsoft.com/en-us/learn/
- We already have far too much stuff on the header, on a mobile phone this will soon take half of the screen

PierreBesson:
Yes this is how Microsoft and Google do it but it's awful ! Especially as they always redirect me to the French version, even though my locale is en_FR. The current position is almost impossible to find.

PierreBesson:
This is how it looks right now in the chinese website. The important thing is that any page can be switched to english.

Screenshot 2019-05-03 at 13 31 45
Screenshot 2019-05-03 at 13 31 27

PR #731, opened 2019-05-06
Hide generator-jhipster-{module|blueprint} from marketplace

murdos:
They're not module or blueprint per say, they're generators of module and blueprint


jdubois:
Thanks @murdos !

PR #734, opened 2019-05-19
add documentation for jacoco exclusion

atomfrede:
Updates https://github.com/jhipster/generator-jhipster/issues/9631

Still need to look into the maven part, but it seems to work similar. Exclude classes from agent analysis and from reporting too.


jdubois:
Excellent, thanks @atomfrede

PR #735, opened 2019-05-20
Traefik : fix new path introduce in 6.0.0

antarus:
Fix default path,  add in 6.0.0 for traefik.


jdubois:
Oh I forgot about that, thanks so much @antarus !

PR #736, opened 2019-05-21
add documentation for jacoco exclusion via maven

atomfrede:
Same as https://github.com/jhipster/jhipster.github.io/pull/734 but for maven.


jdubois:
Thanks @atomfrede !

PR #742, opened 2019-05-28
Add Yeoman install for blueprint in documentation

nonomoho:


pascalgrimaud:
Thanks @nonomoho and @hdurix

Indeed, we need yeoman to use blueprint. 
See https://github.com/jhipster/jhipster-vuejs/blob/master/test-integration/scripts/10-install-jhipster-vuejs.sh#L87-L90
But I agree, it was missing in our doc

PR #743, opened 2019-05-28
Fix missing html <p> tags

avdev4j:
The docker-compose is not rendered properly due to missing tags.

Capture d’écran 2019-05-29 à 00 48 41


pascalgrimaud:
good catch @avdev4j ! thanks

PR #751, opened 2019-06-16
remove gatling 2 support note as we now support gatling 3

atomfrede:
As we now support gatling 3 we should have removed that note already.


pascalgrimaud:
thanks @atomfrede

PR #762, opened 2019-07-13
Better instructions for baseHref when using context path

RickPoleshuck:
There are two minor changes to the documentation based on the normal usage of a production war file will usually use a context path and that normal development does not.


jdubois:
Thanks @RickPoleshuck !

PR #763, opened 2019-07-17
update cassandra integration test documentation

atomfrede:
clearly state docker is now required as we use testcontainers

closes https://github.com/jhipster/generator-jhipster/issues/10052


jdubois:
Awesome!

PR #764, opened 2019-07-19
add explicit clean command to production build commands

atomfrede:
to be consistent with generated readme.

In general doing clean before production build seems reasonable to make sure
everything is build from a clean state imho.

closes https://github.com/jhipster/generator-jhipster/issues/10098


pascalgrimaud:
thanks @atomfrede

PR #773, opened 2019-08-16
Add Matt's UK + Ireland JUG Tour

mraible:


mraible:
I adjusted the font size in the events section. I think it's easier to read when it's smaller. 

Before:

image

After:

image

PR #777, opened 2019-08-28
update gradle command to generate Jacoco xml reports

clement26695:
Update doc following jhipster/generator-jhipster#10293

ping @atomfrede


atomfrede:
:+1:

PR #781, opened 2019-08-31
join climate strike

deepu105:
Discussion here https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/jhipster-dev/tj4sVWbv_Lw

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -483,6 +483,15 @@
         window.top.location.href = "https://www.jhipster.tech";
     }
     </script>
+    <!-- Climate strike JS -->
+    <script type="text/javascript">
+        var DIGITAL_CLIMATE_STRIKE_OPTIONS = {
+          disableGoogleAnalytics: true, // @type {boolean}
+          showCloseButtonOnFullPageWidget: true, // @type {boolean}
+        };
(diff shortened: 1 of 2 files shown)


jdubois:
- I've just tested it and it works fine today. But how will it work on September, 20th, is there a way to test this?
- We should not merge it if we do a release before September, 20th, or people will have the banner forever in the archive - I don't think we have anything planned, but I feel like it's a bit dangerous to merge it too soon. @deepu105 @pascalgrimaud what do you think about this?

deepu105:
@jdubois there is param which when enabled will show the actual screen. I have tested that and works well. The code automatically switches to that view on September 20

But yes we can merge this few days before 20th

jdubois:
OK, then let's merge this a few days before. Perfect!

pascalgrimaud:
Agree, feel free to merge this

deepu105:
@jdubois @pascalgrimaud lets merge this after 6.3.0 release, I thinking of mentioning about it in my codeOne talk next week

jdubois:
Yes +1

pascalgrimaud:
+1

PR #783, opened 2019-09-02
Update Information About GAE

SudharakaP:
This updates information about the Google App Engine generator, since we updated it to Java 11.

Related to issue: https://github.com/jhipster/generator-jhipster/pull/10284


jdubois:
Thanks a lot @SudharakaP !

PR #784, opened 2019-09-03
mention archunit in documentation

atomfrede:
and add logo. Furthermore I added `target=_blank` to external links, should we do that for all external links in general?

EDIT: Of course this should be merged with the next release when arch unit support is released


jdubois:
Thanks! Merging as next release is today!

PR #785, opened 2019-09-09
Update jdl.md

Aiksyuan:
fix the types supported in the JDL，UUID  does not support Common databases


murdos:
In fact UUID type is supported for all databases since version 6.1.0 and https://github.com/jhipster/generator-jhipster/issues/9793 so the documentation is correct

PR #789, opened 2019-09-17
Flat Admin Theme Repository URL Update

SudharakaP:
Not sure exactly but seems like the original author has deleted this repository; I updated the link to a duplicate


pascalgrimaud:
thanks

PR #791, opened 2019-09-17
Add Blob to Promote App Password instead of Gmail Password

SudharakaP:
Currently we promote users to use their gmail password in the gmail configuration section. However this is a bit insecure and does not enable them to use two factor authentication. I've added a blob point to the information about using an app password instead.


jdubois:
Thanks @SudharakaP !

PR #793, opened 2019-09-17
Promote using Dependabot + Add information on Security Vulnerability Reporting

SudharakaP:
I've recommended using dependency analyzers (such as Dependabot) and also changed information on how to report a vulnerability now that we have Github Security Advisories. 

@jdubois : I think this is what you meant in the following comment; https://github.com/jhipster/generator-jhipster/issues/10401#issuecomment-531448583 ?


jdubois:
Oh we have a security policy :-P
Thanks so much for all your help @SudharakaP

PR #796, opened 2019-09-19
Edit Introduction of Microservices

SudharakaP:
Remove "Angular" since we use different front-ends not only Angular.


pascalgrimaud:
Good catch

jdubois:
Awesome! At some point we'll also need to remove Spring Boot :-)

PR #812, opened 2019-10-02
Update in index.html

Tonterias:
Julien,

I added the two lines we talked about. Let see if it works until year end.

Thanks again.


index.html
```diff
@@ -479,6 +479,10 @@ <h3>New books</h3>
                             <li><a href="https://www.infoq.com/minibooks/jhipster-mini-book-5" target="_blank">The JHipster Mini-Book 5.0</a> by Matt Raible. This edition includes new sections on progressive web apps (PWA), code quality, and securing user data.</li>
                             <li><b>Full Stack Development with JHipster</b> by Deepu K Sasidharan and Sendil Kumar. Get it on <a href="https://www.packtpub.com/application-development/full-stack-development-jhipster" target="_blank">Packt</a> and <a href="https://www.amazon.com/Stack-Development-JHipster-Deepu-Sasidharan/dp/178847631X" target="_blank">Amazon</a>.</li>
                         </ul>
+                        <h3>Free Collaborative learning</h3>
+			            <ul>
+                            <li>Learn by doing with <a href="www.spingular.com" target="_blank">Spingular</a>. Join a live collaborative project made with Jhipster!: Spingular Chat w/Websockets</li>
```
mraible: This URL won't work. It should be https://www.spingular.com.

Tonterias: Thanks Mat,

I have changed it. It is ok now?

PR #820, opened 2019-10-08
Update Languages

SudharakaP:
Update our language support section with new languages recently added

### DIFF
--- a/pages/installing_new_languages.md
+++ b/pages/installing_new_languages.md
@@ -26,7 +26,7 @@ These are the currently supported languages
 *   Albanian
 *   Arabic (Libya)
 *   Armenian
-*   Belarussian
+*   Belarusian
 *   Bengali
 *   Catalan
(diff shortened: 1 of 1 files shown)


pages/installing_new_languages.md
```diff
@@ -26,46 +26,47 @@ These are the currently supported languages
 *   Albanian
 *   Arabic (Libya)
 *   Armenian
-*   Belarusian
+*   Belorussian
```
deepu105: This is not correct, it is called `Belarusian`

SudharakaP: [Belarusian is also spelled as Belorussian](https://en.wikipedia.org/wiki/Belarusian_language#Names). :smile:

deepu105:
Not really the language. Belarusian is the language and Belorussian is the
people. Trust me I have a Belorussian friend :P

Thanks & Regards,
Deepu

PR #776, opened 2019-08-24
Fix to remove deprecation warning at startup

SudharakaP:
Fix for the deprecation warning at startup; ```The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.```

PR #823, opened 2019-10-13
Updated the config file to remove deprecation notice

MathieuAA:


ruddell:
This caused issues changing it in the past.  Might be a good idea to comment this since a PR has been opened 3+ times.

See https://github.com/jhipster/jhipster.github.io/commit/06916fedb29c784168521505d98ff457a4b9b341#diff-aeb42283af8ef8e9da40ededd3ae2ab2 and https://github.com/jhipster/jhipster.github.io/pull/776

MathieuAA:
Ahah! very nice! I'll close it then, and yed: adding a comment may be a good idea :)

PR #824, opened 2019-10-13
Update jdl doc

MathieuAA:
Fixes https://github.com/jhipster/generator-jhipster/issues/10520, fixes https://github.com/jhipster/jhipster-core/issues/382 and fixes https://github.com/jhipster/jhipster-core/issues/372

### DIFF
--- a/pages/jdl.md
+++ b/pages/jdl.md
@@ -310,9 +310,8 @@ entity A {
   myString String required minlength(1) maxlength(42) pattern(/[A-Z]+/)
 }
 ```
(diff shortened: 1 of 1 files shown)


pages/jdl.md
```diff
@@ -300,11 +312,13 @@ entity A {
 ```
 If you're using the generator prior to v4.9.X, you'd need to use patterns like this `pattern('[A-Z]+'`).
```
murdos: Maybe we can remove this now, JHipster 4.9 is quite old now. What do you think?

MathieuAA: Good point, I'll update the PR

PR #826, opened 2019-10-14
Tip: improve developer experience if opening only front end in the IDE

kaidohallik:
closes jhipster/generator-jhipster#10502


pascalgrimaud:
thanks @kaidohallik

PR #827, opened 2019-10-16
add github actions logo in ci/cd section

atomfrede:
Following https://github.com/jhipster/generator-jhipster/pull/10606 and https://github.com/jhipster/generator-jhipster/issues/10599 we should have it in our ci/cd section I would say.

image


pascalgrimaud:
thanks @atomfrede
It will be for the next release

PR #828, opened 2019-10-18
Describe Babel solution to make fix IE11

anthony-o:
Better tip for jhipster/generator-jhipster#10184 issue describing full solution in order to use Babel to make IE11 correctly working with JHipster generated applications.


pascalgrimaud:
thanks @anthony-o

PR #840, opened 2019-10-31
Added more info to the JDL export sub-section

MathieuAA:
It didn't tell how the exported JDL file's name was chosen before, now it does.


pascalgrimaud:
Thanks @MathieuAA

PR #848, opened 2019-11-11
add azure spring cloud to deployment options

atomfrede:
Add azure spring cloud to deployment options on the homepage. @jdubois I hope this is the correct logo. So you have a svg version maybe? Just found a non colored version.

image


jdubois:
Yes it's the good one, thanks a lot!!

PR #849, opened 2019-11-14
Added the new --skip-fake-data when creating an app

leonardortlima:


pascalgrimaud:
thanks @leonardortlima and congrats for your 1st contribution
cc @murdos : I think it's correct

PR #852, opened 2019-11-27
Some more fixes to language proposed by alex

atomfrede:
Still some files to adapt, but this is is a start.


pascalgrimaud:
thanks for the review @murdos
thanks for the PR @atomfrede

PR #859, opened 2019-12-03
remove hint about excluding node_modules

atomfrede:
This PR removes the hint of excluding `node_modules` from indexing in idea as this prevents code completion etc. in frontend files. See https://github.com/jhipster/generator-jhipster/issues/10860 for details

### DIFF
--- a/pages/configuring_ide_idea.md
+++ b/pages/configuring_ide_idea.md
@@ -18,20 +18,6 @@ sitemap:
 
 If you want more control on your setup, you can also choose "Import project".
 
-## Exclude directories
-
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
We need to be carefull about that, as not excluding `node_modules`  will slow down the demo / talk we can do.

Then, if you can develop the front part with idea, it means you have a license.
So for our users who use Community Edition of idea (which is the majority I think), it is not possible.

atomfrede:
Good point, you're right, frontend is only ultimate. Let me rephrase the paragraph.

pascalgrimaud:
In my opinion:
- idea ultimate
    - demo -> node_modules should be excluded to speed up idea
    - real project -> depending on your use (some user still prefer VScode for the front)
- idea community -> should always be excluded

atomfrede:
I will close this and rephrase the paragraph about using idea to code frontend code.

PR #860, opened 2019-12-04
Add modelio-jdl-studio documentation

ambpro:
This PR provides the modelio-jdl-studio module documentation.

modelio-jdl-studio


deepu105:
Hi @ambpro first of all great work and thanks for your interest, but unfortunately we cannot have documentation for third party tools within our official docs unless its a module. The only place where we can list third party modules are here https://www.jhipster.tech/modules/marketplace/#/list

The concern is that we have no control over this module and making it part of official docs would mean we endorse it and we cannot take responsibility for something that is not within our control.

Now we have few options here
1. Make a dummy npm module with this doc as the readme and publish it as  JHipster module so that it gets automatically listed in https://www.jhipster.tech/modules/marketplace/#/list
2. We add a new section for third party tools which are not supported directly by the core team and we add this there

Again thanks for submitting this but let us discuss and find a solution and I hope you understand our concern.

@jhipster/developers @jdubois @pascalgrimaud WDYT?

deepu105:
Btw, I would prefer option 2, where we clearly state that this is third party and hence supported by the author of the tool

pascalgrimaud:
Same as you @deepu105, I'd prefer the option 2.
Thanks a lot for all this work

ambpro:
Hello @deepu105 @pascalgrimaud, 
Thank you for your reply.

> We add a new section for third party tools which are not supported directly by the core team and we add this there

By the option 2, would you mean that we have to create a new section `Third-Party Tools` alongside to the `Tools` section or it's already exist somewhere? Could you please precise a bit more the this point?

avdev4j:
This ticket is opened for 2 years. I can see some documentation about Modelio here: https://www.jhipster.tech/jhipster-uml/#modelioexample

Do you think it's enough, so we can close this one?

MathieuAA:
@avdev4j JUML is no longer maintained. However I agree with the fact that this issue has been open for way too long.
If there's no real consensus here before the end of the week, I propose to close this issue.

PR #861, opened 2019-12-05
Add Correct Paperowls Image

SudharakaP:
@jdubois : I think the recent Paperowl image addition somehow put a wrong logo. So I've corrected it here. :smile:


jdubois:
Thanks! But I put the one from Opencollective so I'm not sure

vishal423:
I think @SudharakaP is right. Current image seems copy of another sponsor `Edusson`

jdubois:
Yes but that's what they put: https://opencollective.com/papersowl-com

jdubois:
Or there's a bug with Opencollective?

SudharakaP:
@jdubois : I've sent them a support email explaining the matter and asking clarification about the correct logo. Will see what they say. :smile:

vishal423:
I think you would like to add https://opencollective.com/papersowl

SudharakaP:
@jdubois @vishal423 : So I've done some investigation. :laughing: :laughing: 

There's two paperowls both of which are bronze sponsors; 

1) https://opencollective.com/papersowl-com

2) https://opencollective.com/papersowl#section-contributions

If you go to each of the above links and click on the website link (below the logo) you'll see they both go to the same site; [https://papersowl.com/write-my-paper-for-me](https://papersowl.com/write-my-paper-for-me). So I think the image proposed in this pr is correct. But I don't understand why one would give two bronze sponsorships rather than combining them. :smile:

jdubois:
To be honnest @SudharakaP I'm guessing this all the same company, and they are trying to have their logo there several times in order to increase their visbility. It's a way to "hack" the sponsoring system - not very nice of them, but it's hard to enforce a rule here, as they act as different companies. For this reason I wouldn't spend too much effort on this.

PR #862, opened 2019-12-07
Make JDL more visible in docs

deepu105:
SInce JDL is the integer part of JHipster and since we want to make it more integral we should make it more visible in docs instead of burying it under tools.

WDYT @jdubois @pascalgrimaud @MathieuAA ?


pascalgrimaud:
💯 % agree, let's do it

MathieuAA:
Thank you Deepu

PR #864, opened 2019-12-12
Azure cloud documentation

jdubois:
Totally re-write the Azure documentation, to detail the usage of the two new Azure sub-generators


pascalgrimaud:
@jdubois : is it for the next release, or it is to update our current documentation ?

jdubois:
It will be for the next release, so it's better to wait before merging it

pascalgrimaud:
Arg I can't merge it as all the file is in conflict :(

PR #882, opened 2020-01-21
remove redirect from /security to /using-uaa/

ruddell:
If you go to https://jhipster.tech/security it redirects to https://jhipster.tech/using-uaa/ instead of https://jhipster.tech/security/ due to the incorrect `redirect_from` config.

Because the `using-uaa` page is from after the conversion from `.html` -> folder paths, I don't think it is needed.


pascalgrimaud:
thanks a lot @ruddell

PR #883, opened 2020-01-21
Update separating-front-end-and-api.md to clarify `jhipster --skip-server` instructions

keithagroves:
`jhipster --skip-server` doesn't work alone and requires arguments. This isn't clear from the instructions on separating the front end.


pascalgrimaud:
Thanks @keithagroves

PR #886, opened 2020-01-25
update code formatting

vivekmore:
use yaml code block


pascalgrimaud:
Thanks but I don't think it will change anything.
Let's merge this anyway

PR #919, opened 2020-03-23
Add Information About Heroku Account Verification

SudharakaP:
While working on https://github.com/jhipster/generator-jhipster/issues/11339, I noticed that if the Heroku account is not verified the build fails due to the inability to install add-ons (as highlighted in this [StackOverflow issue](https://stackoverflow.com/questions/44518035/jhipster-url-must-start-with-jdbc-error-on-heroku/44530428#44530428 )). I think we need to add this to the documentation as it's difficult to figure out from the build failure logs the reason. 

Related to https://github.com/jhipster/generator-jhipster/issues/11339

### DIFF
--- a/pages/heroku.md
+++ b/pages/heroku.md
@@ -39,6 +39,10 @@ To deploy your application to Heroku, run this command:
 
 This should package your application in "production" mode, create an Heroku application with a database, upload your code, and start the application.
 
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
thanks @SudharakaP

SudharakaP:
You are welcome :smile:

PR #926, opened 2020-03-30
Adding examples for Angular.

amatosg:
This could help new users to understand how filtering works using Angular.


amatosg:
@DanielFran I made the changes, as I can read from the PR, the use of notIn should be almost the same as `in`

DanielFran:
@amatosg I was thinking that you can join in/notin, equals/notequals in same example like you do for less/greater... Just to be consistent.

amatosg:
@DanielFran you're right, it should be grouped.

PR #937, opened 2020-04-14
Blueprint phase steps ordering

danielpetisme:
Doc proposal for:
https://github.com/jhipster/generator-jhipster/issues/11577

### DIFF
--- a/modules/creating_a_blueprint.md
+++ b/modules/creating_a_blueprint.md
@@ -145,6 +145,34 @@ There are multiple ways to customize a phase from JHipster.
     }
 ```
 
+1) Re-order the phase steps, this is when the blueprint gets the phase from JHipster and generates a new step sequence.
+
+```javascript
+    // Run the blueprint steps before any parent steps
+    get initializing() {
+        const phaseFromJHipster = 
+        return {
+            myCustomInitPhaseStep() {
+                // Do all your stuff here
+            },
+            ...super._initializing();
+        }
+    }
+    // Fine control of the steps order
+    get configuring() {
+        const phaseFromJHipster = super._configuring();
+        return {
+            insight: phaseFromJHipster.insight,
+            configureGlobal: phaseFromJHipster.configureGlobal,
+            myCustomConfigPhaseStep() {
+                // Do all your stuff here
+            },
+            saveConfig: phaseFromJHipster.saveConfig
+
+        }
+    }
+```
+
 You can also access to JHipster's variables and functions directly from a Blueprint.
 
 ## Available variables and functions
--- a/modules/creating_a_blueprint.md
+++ b/modules/creating_a_blueprint.md
@@ -145,7 +145,7 @@ There are multiple ways to customize a phase from JHipster.
     }
 ```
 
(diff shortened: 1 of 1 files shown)


modules/creating_a_blueprint.md
```diff
@@ -145,6 +145,34 @@ There are multiple ways to customize a phase from JHipster.
...
+                // Do all your stuff here
+            },
+            ...super._initializing();
+        }
+    }
+    // Fine control of the steps order
+    get configuring() {
+        const phaseFromJHipster = super._configuring();
+        return {
+            insight: phaseFromJHipster.insight,
```
mshima, line 165: The problem with this approach is that if generator-jhipster adds a new step, the blueprint will fail and it will be difficult the find the problem.
Maybe add a verification if the result has every key from jhipsterStep.

danielpetisme: Something like 
```
for (step in phaseFromJHipster) {
  customPhase.hasOwnProperty(step)
}
```
In one hand, I undertand the additional verification but it the other it looks tricky.
Overall re-ordering steps implies a strong coupling between the upstream generator and the blueprint.
Maybe just adding a warning would be enough:
> While finely control the phase steps gives you a great flexibility, it's  error prone. If generator-jhipster adds a new step, the blueprint will fail and it will be difficult the find the problem. Use it at your own risk.

WDYT?

mshima: Sure.

danielpetisme: I reconsider my position. I don't think a fine control on steps order is something to promote. As you said @mshima it's risky and hard to troubleshoot. I think a basic decoration (pre/post phase) should do the job.

danielpetisme:
@deepu105 any additional comment or I can merge it?

modules/creating_a_blueprint.md
```diff
@@ -145,6 +145,29 @@ There are multiple ways to customize a phase from JHipster.
...
+        const customPostPhaseSteps = {
+            myCustomPostInitStep() {
+                // Stuff to do AFTER the JHipster steps
+            }
+        };
+        return {
+            ...customPrePhaseSteps,
+            ...super._initializing(),
+            ...customPostPhaseSteps,
+        }
```
mshima, line 167:
```suggestion
        };
```

modules/creating_a_blueprint.md
```diff
@@ -145,6 +145,29 @@ There are multiple ways to customize a phase from JHipster.
...
+        };
+        const customPostPhaseSteps = {
+            myCustomPostInitStep() {
+                // Stuff to do AFTER the JHipster steps
+            }
+        };
+        return {
+            ...customPrePhaseSteps,
+            ...super._initializing(),
+            ...customPostPhaseSteps,
```
mshima, line 166:
```suggestion
            ...customPostPhaseSteps
```

deepu105:
Haven't looked at it yet, let me see

deepu105:
Ok it looks good

PR #946, opened 2020-04-25
Altering Documentation to Add MongoDB Relationship Support

SudharakaP:
This make some small changes to our documentation in order to reflect work done on MongoDB side to support relationships. Please let me know if anything needs to be changed or if there's other places in documentation we need to change. 

Related to https://github.com/jhipster/generator-jhipster/issues/7944


pascalgrimaud:
good catch @SudharakaP

SudharakaP:
@pascalgrimaud : Actually it was @ruddell  who pointed that out through our Gitter channel. :smile:

PR #951, opened 2020-04-28
skipUserManagement should be documented `false`

vanthuanqs:
According to this document:
https://www.jhipster.tech/creating-an-app/
> --skip-user-management - Skip the user management generation, both on the back-end and on the front-end (Default: false)

the skipUserManagement config should be documented as `false` instead `true`


atomfrede:
Thanks @vanthuanqs and congrats to your first contribution. You are right `skipUserManagement` is `false` by default.

PR #952, opened 2020-04-28
Add Micronaut and Quarkus samples

mraible:


mraible:
Hmmmm, I just noticed this repo has no "issues" tab. I think it'd be useful for folks requesting clarification or logging a request. For example, related to this, I wanted to create an issue that says to add all our new backends and their logos to the "Server Side Options" section on the site.

jdubois:
Yes this is to have everything in jhipster /generator-jhipster

mraible:
Good to know! I added an issue to add them.

PR #955, opened 2020-05-02
Add Tip About Time Drift in Docker

SudharakaP:
This adds a tip about the time drift that can occur in docker containers when sleeping for extended periods of time.

Closes https://github.com/jhipster/generator-jhipster/issues/11659


SudharakaP:
> Docker should be capitalized. slightly_smiling_face

Done :+1:

PR #956, opened 2020-05-02
JDL: reorder entity types to group dates and blobs

murdos:
Instant was at the end of the table, it's more logic to have it besides LocalDate, ZonedDateTime and Duration.
Ditto for ByteBuffer: it should be next to blob types


atomfrede:
I recently was thinking the same and wondering why I can't find instant until I realized its further down :+1:

PR #957, opened 2020-05-03
add doc on running sql tests with testcontainers

ecostanzi:
Related to: https://github.com/jhipster/generator-jhipster/issues/11544


atomfrede:
@ecostanzi Maybe we can add the testcontainer logo to the frontpage like we do for other things

ecostanzi:
Sure!

PR #964, opened 2020-05-21
remove non existing downloads property from marketplace items

atomfrede:
It seems the `downloads` property does not exist in the npm api anymore.
This pr removes the broken display and uses order by popularity (instead of the final score, which leads to little changes in the order).

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -1007,7 +1007,7 @@ sitemap:
                 <br>
             </div>
             <div class="row">
-                <div ng-cloak ng-repeat="module in modules | filter: searchText | orderBy: '-downloads' | limitTo:16">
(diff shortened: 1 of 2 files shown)


ruddell:
The npm API does not support the lookup of downloads for scoped NPM modules. `@xiges/generator-jhipster-siddhi` breaks the requests.  You can see the 400 error in the browser console on https://www.jhipster.tech/

Works: https://api.npmjs.org/downloads/point/last-month/generator-jhipster-vuejs,ignite-jhipster
Broken: https://api.npmjs.org/downloads/point/last-month/generator-jhipster-vuejs,@xiges/generator-jhipster-siddhi

This seems like a good fix to me though 👍

PR #965, opened 2020-05-23
added link to "options" sub-page

62mkv:


62mkv:
closes #11732

62mkv:
@MathieuAA would it make sense to also re-structure "Options" page, to display entity-specific options separately from field-specific ones? not sure how big is an overlap

MathieuAA:
What do you mean by field-specific options? Are you referring to the validations?

pages/jdl/entities.md
```diff
@@ -296,3 +296,5 @@ Common databases:
     <td><dfn>required, minbytes, maxbytes, unique</dfn></td>
   </tr>
 </table>
+
+[Options]: ./options.html "Options"
```
MathieuAA, line 300: Simpler: "[Options]: options "Options"

62mkv:
nevermind, I noticed that the title above the table says that those specifically *entity* options, so it's fine

62mkv:
anything else here?

PR #967, opened 2020-05-29
switch to registry.npmjs.org instead of api

atomfrede:
As suggested on the mailinglist we use registry.npmjs.org instead of api.npmjs.io. This has a bit changed order, but returns e.g. micronaut blueprint and seems to return more recent versions (e.g. vuejs 1.8.0 while 1.8.1 is the latest release).

### DIFF
--- a/js/services.js
+++ b/js/services.js
@@ -66,7 +66,7 @@
             },
             getAllModules: function (start, size) {
                 /* Get all Jhipster modules */
-                return $http.get('https://api.npms.io/v2/search?q=keywords%3Ajhipster-module%2Cjhipster-blueprint&from=' + start + '&size=' + size).success(function (resp) {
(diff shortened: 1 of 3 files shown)


js/services.js
```diff
@@ -66,7 +66,7 @@
             },
             getAllModules: function (start, size) {
                 /* Get all Jhipster modules */
-                return $http.get('https://api.npms.io/v2/search?q=keywords%3Ajhipster-module%2Cjhipster-blueprint&from=' + start + '&size=' + size).success(function (resp) {
+                return $http.get('http://registry.npmjs.org/-/v1/search?text=keywords:jhipster-blueprint,jhipster-module&from=' + start + '&size=' + size).success(function (resp) {
```
mshima, line 69:
```suggestion
                return $http.get('https://registry.npmjs.org/-/v1/search?text=keywords:jhipster-blueprint,jhipster-module&from=' + start + '&size=' + size).success(function (resp) {
```

PR #978, opened 2020-07-06
Fix package-lock.json link format

bagage:
Currently looking bad:

image


pascalgrimaud:
Thanks and congrats for your 1st contribution @bagage

PR #980, opened 2020-07-06
minor cleanup

vivekmore:


_config.yml
```diff
@@ -1,17 +1,18 @@
-safe:        true
+safe: true
 markdown: kramdown
 exclude: [node_modules, bower_components, vendor]
 url: https://www.jhipster.tech
 github_repository_url: https://github.com/jhipster/jhipster.github.io
 kramdown:
   input: GFM
-gems:
+plugins:
```
ruddell, line 8: This change has caused issues in the past, not sure how to reproduce it though.  See https://github.com/jhipster/jhipster.github.io/pull/557 and https://github.com/jhipster/jhipster.github.io/pull/776

vivekmore: Understood, I'll revert this.
Should I add a note/comment so another such attempt is avoided?

vivekmore: done and done! ✔️

vivekmore:
@ruddell - is there anything else you'd like me to address?

_config.yml
ruddell: I don't the change or comment.  I still see `gems` being changed to `plugins`, am I missing the change?

Other spacing/renamings look OK to me, did not test locally.

vivekmore: I had reverted it, but it looks like I goofed up in one of the rebases since ... (or I never reverted it - which is also likely 😨)

vivekmore: I've reverted it now

pascalgrimaud:
Thanks a lot @vivekmore

PR #987, opened 2020-07-20
update context path documentation for swagger usage

atomfrede:
closes https://github.com/jhipster/generator-jhipster/issues/12073

### DIFF
--- a/pages/production.md
+++ b/pages/production.md
@@ -175,6 +175,17 @@ There are many other options that you can find in [Spring Boot documentation](ht
 
 When deploying a JHipster app to an Application Server or customizing your context-path, it is required to set the `base` value in `webpack.common.js`or `webpack.prod.js` equal to the expected context-path.
 
+In order to use the swagger api documentation site you need to adapt the swagger index file under `src/main/webapp/swagger-ui/index.html`. 
+Assuming your context path is `myContextPath` you need to modify `index.html` like
+
+```
+response.data.forEach(function (resource) {
+    urls.push({"name": resource.name, "url": "/myContextPath" + resource.location});
+});
+```
+
+See [this issue for details](https://github.com/jhipster/generator-jhipster/issues/12073).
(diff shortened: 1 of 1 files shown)


pages/production.md
```diff
@@ -175,6 +175,17 @@ There are many other options that you can find in [Spring Boot documentation](ht
 
 When deploying a JHipster app to an Application Server or customizing your context-path, it is required to set the `base` value in `webpack.common.js`or `webpack.prod.js` equal to the expected context-path.
 
+In order to use the swagger api documentation site you need to adapt the swagger index file under `src/main/webapp/swagger-ui/index.html`. 
+Assuming your context path is `myContextPath` you need to modify `index.html` like
+
+```
+response.data.forEach(function (resource) {
```
pascalgrimaud, line 182: can you add the axios get part too plz ?

pages/production.md
```diff
@@ -175,6 +175,17 @@ There are many other options that you can find in [Spring Boot documentation](ht
 
 When deploying a JHipster app to an Application Server or customizing your context-path, it is required to set the `base` value in `webpack.common.js`or `webpack.prod.js` equal to the expected context-path.
 
+In order to use the swagger api documentation site you need to adapt the swagger index file under `src/main/webapp/swagger-ui/index.html`.
```
vishal423, line 178: We also have similar details under `Building an executable JAR / WAR file`. It would be good to club along that.

atomfrede: Maybe we can add the snippet (also it is duplicated) for webpack configuration here too.

vishal423: I think the `Build` section is more appropriate for such customization and should remove all duplication from here. WDYT?

atomfrede: Good point not sure what people are looking for. I will have a look and see if removing the dedicated running under context path makes sense eventually.

PR #994, opened 2020-08-10
Add Heroku + Docker documentation

mraible:
Depends on https://github.com/jhipster/generator-jhipster/pull/12237.

### DIFF
--- a/pages/heroku.md
+++ b/pages/heroku.md
@@ -202,12 +202,18 @@ When you use `jhipster heroku` to deploy your application, it properly configure
 heroku config:get DATABASE_URL --remote docker
 ```
 
+This command will retrieve a value with the following syntax:
+
+```
+postgres://username:password@address
+```
+
 Then, set the database environment variables to match the keys that are in `application-heroku.yml`:
 
 ```shell
-heroku config:set JDBC_DATABASE_URL=<value-after-@-from-last-command> --remote docker
-heroku config:set JDBC_DATABASE_USERNAME=<username-value-from-last-command> --remote docker
-heroku config:set JDBC_DATABASE_PASSWORD=<password-value-from-last-command> --remote docker
+heroku config:set JDBC_DATABASE_URL=jdbc:postgresql://<address> --remote docker
+heroku config:set JDBC_DATABASE_USERNAME=<username> --remote docker
+heroku config:set JDBC_DATABASE_PASSWORD=<password> --remote docker
 ```
 
 Set the max amount of Java memory to use and specify the Spring profiles.
--- a/pages/heroku.md
+++ b/pages/heroku.md
@@ -135,6 +135,13 @@ Or when using gradle:
 
 You can deploy your app as a Docker container to Heroku too. While this works, there's no Heroku setup and configuration that happens, so you have to do that manually. This documentation assumes you've already run `jhipster heroku` to deploy your app and therefore leverages the integration and add-on provisioning that this process performs.
 
+**NOTE**: If you're using a version of JHipster that's prior to v6.10.2, you'll need to add the following to `src/main/resources/config/application-heroku.yml`:
+
+```yaml
+server:
+  port: ${PORT:8080}
+```
+
 Build your Docker image:
 
 ```
--- a/pages/heroku.md
+++ b/pages/heroku.md
@@ -238,7 +238,7 @@ Now you should be able to open your app, click the **sign in** link, and authent
 heroku open --remote docker
 ```
 
-If you test your Dockerfied JHipster app on [securityheaders.com](https://securityheaders.com), you'll see it scores an **A**!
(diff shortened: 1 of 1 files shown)


pages/heroku.md
```diff
@@ -131,6 +131,115 @@ Or when using gradle:
...
+heroku logs --tail --remote docker
+```
+
+Now you should be able to open your app, click the **sign in** link, and authenticate!
+
+```
+heroku open --remote docker
+```
+
+If you test your Dockerfied JHipster app on [securityheaders.com](https://securityheaders.com), you'll see it scores an **A**!
```
SudharakaP, line 241: Think the word commonly used is `Dockerized` ?

pages/heroku.md
```diff
@@ -131,6 +131,115 @@ Or when using gradle:
...
+heroku config:set JDBC_DATABASE_URL=<value-after-@-from-last-command> --remote docker
+heroku config:set JDBC_DATABASE_USERNAME=<username-value-from-last-command> --remote docker
+heroku config:set JDBC_DATABASE_PASSWORD=<password-value-from-last-command> --remote docker
+```
+
+Set the max amount of Java memory to use and specify the Spring profiles.
+
+```shell
+heroku config:set JAVA_OPTS=-Xmx256m
+heroku config:set SPRING_PROFILES_ACTIVE=prod,heroku,no-liquibase
```
atomfrede, line 210: I am not sure you should set `no-liquibase` here. If there are new changes and you don't deploy it via git/jar before liquibase won't run with the docker container (as it is run during release hook via dedicated mvn command). For gradle we need the liquibase part in any case I think.

pages/heroku.md
```diff
@@ -131,6 +131,115 @@ Or when using gradle:
...
+
+Run the command below to open your browser and navigate to your app.
+
+```
+heroku open --remote docker
+```
+
+Copy the URL of your app and log in to your Okta developer account. Go to **Applications** > **Web** > **General** and add the URL to Login and Logout redirect URIs. Make sure the login redirect URI ends with `/login/oauth2/code/oidc`.
+
+NOTE: You will **NOT** be able to login to your JHipster app using the admin account the Okta add-on provisions. If you receive a stack trace when logging in, log out and try again with the account you created when you first deployed to Heroku.
```
atomfrede, line 221: Maybe something like `to make sure you are not logged in anymore either log out or use a new private window` makes it more clear?

pages/heroku.md
mraible: OK, I'll remove it.

pages/heroku.md
```diff
@@ -131,6 +131,115 @@ Or when using gradle:
...
+
+```
+heroku config:get DATABASE_URL --remote docker
+```
+
+Then, set the database environment variables to match the keys that are in `application-heroku.yml`:
+
+```shell
+heroku config:set JDBC_DATABASE_URL=<value-after-@-from-last-command> --remote docker
+heroku config:set JDBC_DATABASE_USERNAME=<username-value-from-last-command> --remote docker
```
atomfrede, line 202: Not exactly sure here if I understand what you mean with `username-value-from-last-command`

pages/heroku.md
```diff
@@ -131,6 +131,115 @@ Or when using gradle:
 
 `heroku deploy:jar build/libs/*jar`
 
+## Deploying Docker to Heroku
+
+You can deploy your app as a Docker container to Heroku too. While this works, there's no Heroku setup and configuration that happens, so you have to do that manually. This documentation assumes you've already run `jhipster heroku` to deploy your app and therefore leverages the integration and add-on provisioning that this process performs.
+
+Build your Docker image:
```
mraible, line 138: @atomfrede Do you think I should include [these changes](https://github.com/jhipster/generator-jhipster/pull/12237) as a note for developers using JHipster < 6.10.2?

pages/heroku.md
mraible: The `heroku config:get DATABASE_URL` returns something like the following:

```
postgres://gesofkxyjv3umrel:87e2b57d83efe22595as2f11b7cf6195fd28f3e767e7fd39071ea092853fe4ed3@ec2-52-204-20-42.compute-1.amazonaws.com:5432/dag2k3tic441if
```

I expect that developers will know `//username:password@` syntax, but maybe I'm wrong?

atomfrede: Agree. It find it a bit confusing because you use `value-after-@-from-last-command` instead of e.g. database address from last command. Maybe make it more clear by doing something like

```
heroku config:get DATABASE_URL returns a string in the form postgres://username:password@address
```
and it would be `<address>`, `<username>`. But not sure it makes it easier.

pages/heroku.md
atomfrede: Yes, guess that would prevent stackoverflow questions or issues on our side.

pages/heroku.md
mraible: Fixed!

PR #995, opened 2020-08-20
Adds the micronaut logo to the website

sdelamo:
This is how it looks like: 

Screenshot 2020-08-20 at 09 19 49
Screenshot 2020-08-20 at 09 23 11


deepu105:
We weren't adding the module related techs here (thats why you don't see
nodejs, .net etc as well, so we need to make decision about that first)

deepu105:
I would rather add module related techs in a separate section from main
generator techs to be clear

sdelamo:
Do you want me to split to a separate PR the Micronaut Logo in "server side options"?

deepu105:
Ya lets add a section below server side tech with title "technologies
available as modules" and add this there, then we can also add nodejs, .net
and quarkus there later

sdelamo:
I think Micronaut should be listed in the current home page "Server side options" section.

sdelamo:
Moreover, since you can use micronaut instead spring boot I think that it is relevant change to the explanation image about what is JHipster 

Screenshot 2020-08-20 at 09 23 11

atomfrede:
> Moreover, since you can use micronaut instead spring boot I think that it is relevant change to the explanation image about what is JHipster
> 
> Screenshot 2020-08-20 at 09 23 11

I agree here somehow. As we have added vue here before it was part of the core (in fact it is still an official blueprint like micronaut). And the text already mentions other service side blueprints. But if we put all logos here it may become a bit cluttered. For now I am totally ok by adding the logo, maybe in the long term we can create a combined image with the relevant logos, so it becomes again `B + F = JH` instead of `B1/B2/B3 + F1/F2/F3 = JH`

deepu105:
Well I'm not against adding logo, I'm just against making a clutter. And if
we are adding official blueprint logos then we should add all of them,
thats only fair.

sdelamo:
@deepu105 

what you are suggesting is that the image of the formula should display?

`Spring Boot or Micronaut or Quarkus or NodeJS or .Net + Angular or React or Vue = JHipster`

deepu105:
Ya but then it would look quite messy. So may be better to avoid that formula graphic altogether

atomfrede:
I would say let's add the micronaut logo to the service side options now and we will find a better solution for the "what is jhipster" formula, wdyt?

deepu105:
Ok

atomfrede:
@pascalgrimaud @jdubois Are you okay with the proposal (adding micronaut and other blueprints to supported options and find nice solution for the what is jhipster formula later) too?

jdubois:
Hey - thanks for mentioning me @atomfrede as I totally missed this (I can't watch all issues, there's just too many of them).

I'm totally fine adding the Micronaut logo in the "server side" options.

Then, for the main "what is JHipster" section, I agree with Deepu that we are going to have a clutter here. There's also Quarkus, and the NestJS blueprint which seem quite good (and we give money to both of them). So at some point it's not going to work.

This "spring+angular=jhipster" was a great logo as long as we only had both of them, and it is now becoming increasingly wrong. Could we do a carousel here? Like we have for companies logos, but with "spring+angular=jhipster", "spring+vuejs=jhipster", "micronaut+angular=jhipster" ....

atomfrede:
Sure we can try. So I would say let's merge it here (I have removed the part from the "spring+angular=jhipster" part. I will take care of adding our other server side options over the weekend.

jdubois:
OK @atomfrede but do you have the tools to do the carousel? Also have a look at how it works for the companies, just copy/paste it as it works fine. I believe it's old AngularJS code I did at the beginning!
By the way, I just had our monthly Cloudflare report, we have more than 100,000 unique visitors per month on this main website, so that's important!

atomfrede:
I guess we find a nice carousel solution here.

PR #1003, opened 2020-09-03
add micronaut family member to artwork page

atomfrede:
This adds the micronaut family member to the artwork page. I can't find the files in the local repository I guess there is a redirect/configuration to our artwork repo somewhere, is that correct?


jdubois:
@atomfrede yes there's a redirect

jdubois:
Argh the size is not the same as the other images:
Screenshot 2020-10-01 at 14 12 19

jdubois:
As the size wasn't correct I reverted it in #1013

atomfrede:
Seen it too 😅

jdubois:
It's just sample images, honestly no need to spend much time on this

PR #1008, opened 2020-09-20
Fix 404 after logging with traefik and oauth2

chelfe:
fix: https://github.com/jhipster/generator-jhipster/issues/11782


pascalgrimaud:
Thanks @chelfe

PR #1015, opened 2020-10-03
Fix: bracket typo

bendeguz-szatmari:
Replace ( with { at line 59


pascalgrimaud:
Thanks @bendeguz-szatmari

PR #1016, opened 2020-10-06
correct entando footer image link

atomfrede:
This PR corrects the entando image in the page footer.

Current state

image

Correct:
image


deepu105:
Oops sorry my bad

PR #1017, opened 2020-10-08
Add instructions for Okta CLI

mraible:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "jhipster-public-wesite",
   "devDependencies": {
-    "browser-sync": "2.26.7",
+    "browser-sync": "^2.26.12",
```
pascalgrimaud, line 4: I think it's better to use "fixed" version. Can you remove `^` plz?

pascalgrimaud (changes requested):

pascalgrimaud (approved):

PR #1028, opened 2020-11-01
Add Christophe Bornet as OpenAPI stream lead

cbornet:


pascalgrimaud:
LGTM, as you're the boss in this part @cbornet ! :-D

PR #1029, opened 2020-11-05
Fixed reflexive relationship example

tgeens:
A reflexive relationship can never have a required side. There was already a note about that limitation, but the example still had the `required` keyword.

See discussion in https://github.com/jhipster/generator-jhipster/issues/11495


pascalgrimaud:
thanks @tgeens

PR #1030, opened 2020-11-09
Remove passages for context path configuration in production build section

clocken:
With [generator-jhipster#12649](https://github.com/jhipster/generator-jhipster/issues/12649) there is no need to configure the SwaggerUI for a context path anymore.
For Angular clients the `base` tag in `index.html` has to be set and for React and Vue clients the `base` property of `HtmlWebpackPlugin` has to be set.


avdev4j:
hi, @clocken is this PR is still relevant? 

wdyt @PierreBesson ?

clocken:
Hi @avdev4j, yes this is still relevant. It should be merged with the corresponding PR's for [generator-jhipster#12649](https://github.com/jhipster/generator-jhipster/issues/12649)

PR #1038, opened 2020-11-27
Add Quarkus as official blueprint

avdev4j:
https://github.com/jhipster/generator-jhipster-quarkus/issues/71

### DIFF
--- a/modules/official-blueprints.md
+++ b/modules/official-blueprints.md
@@ -19,35 +19,35 @@ JHipster supports several official blueprints. These blueprints have two main go
 
 ## Kotlin
 
-First official blueprint, it replaces most of Java back-end code with Kotlin. Source code and documentation could be found here: [repository](https://github.com/jhipster/jhipster-kotlin).
+It replaces most of Java back-end code with Kotlin. Source code and documentation could be found here: [repository](https://github.com/jhipster/jhipster-kotlin).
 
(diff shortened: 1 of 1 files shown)


mraible (approved):

modules/official-blueprints.md
```diff
@@ -37,13 +37,17 @@ Stream lead is: [Daniel Petisme](https://github.com/danielpetisme).
 
 ## Node.js
 
-Latest official blueprint, it replaces the server java side with the NestJS Node framework. Source code and documentation could be found here: [repository](https://github.com/jhipster/generator-jhipster-nodejs).
+Fourth official blueprint, it replaces the server java side with the NestJS Node framework. Source code and documentation could be found here: [repository](https://github.com/jhipster/generator-jhipster-nodejs).
```
joewhite101, line 40: Numbering these probably isn't a good idea. Will be a maintenance headache and doesn't adding anything. I think just a general description for each is good.

Removing latest is a good idea but i don't think the ordering is needed.

joewhite101 (changes requested):
Small change

modules/official-blueprints.md
avdev4j: I wonder the same, to be honest. We should not add the idea there is a kind of order. I'll remove all this numbers.

DanielFran (approved):
LGTM

PR #1040, opened 2020-12-02
Improve Git rename detection hint for manual upgrade

kaidohallik:
Follow up to #1039

I tried manual upgrade where I had ~800 renamed files and Git rename detection started to work after I set `diff.renameLimit=1000`.


kaidohallik:
Converted to draft because trying to add additional information. Entity list component in some configuration has >50% changed comparing to JHipster v6 so for Git rename detection to work for entity list components, may need lowering similarity level.

kaidohallik:
Hint updated, added similarity threshold change information.

PR #1054, opened 2021-01-02
The example code may be wrong in "Do not use  `@MapsId`"

Farley-Chen:
I think the example code in "Do not use `@MapsId`" may be wrong.
If it's a bidirectional one-to-one, may be it should be like what I have edited，like this
```
class Car{ @OneToOne @JoinColumn(name="current_driver_id") Driver currentDriver} // car can be drived by another driver in future
class Driver{@OneToOne(mappedBy = "currentDriver") Car drivingCar} // driver drives another car in future
```
if they're two unidirectional one-to-one, maybe 'id' shouldn't be the join column I think.

I'm not native speaker of English, so please point out if I say something wrong.

Thanks

### DIFF
--- a/pages/managing_relationships.md
+++ b/pages/managing_relationships.md
@@ -422,8 +422,8 @@ Do not use  `@MapsId` when:
     For eg,
 
     ```
-    class Car{ @OneToOne @JoinColumn(name="id") Driver currentDriver} // car can be drived by another driver in future
-    class Driver{@OneToOne @JoinColumn(name="id") Car drivingCar} // driver drives another car in future
+    class Car{ @OneToOne @JoinColumn(name="current_driver_id") Driver currentDriver} // car can be drived by another driver in future
+    class Driver{@OneToOne(mappedBy = "currentDriver") Car drivingCar} // driver drives another car in future
     ```
     Both car and driver association value may change in future.
 


avdev4j:
could we merge this one @pmverma ?

pmverma:
Thanks @Farley-Chen for the correction.
The doc there is just an example to understand the concept.
I agree the missing `mappedBy` but I doubt about `id -> current_driver_id`.
Unless there is a medium table for eg, in case of ManyToMany, jhipster uses only column `id`.

Did you had `current_driver_id` in your liquibase xml file?

Farley-Chen:
@pmverma 

Thanks for replying me.

Yes, there is a column "current_driver_id" in my liquid xml

Here are what I get by this JDL
JDL:
```
entity Driver
entity Car

relationship OneToOne {
  Car{currentDriver} to Driver{drivingCar} 
}
```

liquidbase xml (part of it):
```xml
<changeSet id="20210303155109-1" author="jhipster">
        <createTable tableName="car">
            <column name="id" type="bigint" autoIncrement="${autoIncrement}">
                <constraints primaryKey="true" nullable="false"/>
            </column>
            <column name="current_driver_id" type="bigint">
                <constraints unique="true" nullable="true" uniqueConstraintName="ux_car_current_driver_id" />
            </column>
            <!-- jhipster-needle-liquibase-add-column - JHipster will add columns here -->
        </createTable>
    </changeSet>
```

entity car java (part of it)
```java
public class Car implements Serializable {

    private static final long serialVersionUID = 1L;

    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @OneToOne
    @JoinColumn(unique = true)
    private Driver currentDriver;
...
}
```


Here we don't use `@MapsId`, so a new column seems to be created in `car table` to add ForeignKey constraint

I think for OneToMany or OneToOne, only if using `@MapsId`, column `id` will be used, otherwise will create a join column to add foreign key

mraible:
Should we merge this in @pmverma?

pmverma:
Sure @mraible! I missed the notification as I am not much active on the internet lately.

PR #1057, opened 2021-01-07
Supress unmaintained Heroku Addon

gmarro:
The mLab team has discontinued their MongoDB add-on. The mLab MongoDB add-on will be removed from all Heroku apps on November 10, 2020 and graphenedb addon does not exist anymore.


atomfrede:
Thanks @gmarro for the heads up. We must also adapt the main generator not to support/use mLab and Graphene anymore (which is very sad imho).

atomfrede (changes requested):
There is also a part `Using Neo4j` (under `Troubleshooting`) which we should remove.

atomfrede (approved):

PR #1063, opened 2021-01-14
Update 011_tip_configuring_email_in_jhipster.md

bendeguz-szatmari:
My emails were marked as spam without setting the 'jhipster.mail.from' property properly. Tested with Gmail and Outlook.


pascalgrimaud:
Thanks @bendeguz-szatmari

PR #1064, opened 2021-01-16
correct okta cli usage command

atomfrede:
When trying out okta cli it did not work as described. I only got the following options:

```
➜ okta apps create
Application name [okta-cypress]: 
Type of Application
(The Okta CLI only supports a subset of application types and properties):
> 1: Web
> 2: Single Page App
> 3: Native App (mobile)
> 4: Service (Machine-to-Machine)
Enter your choice [Web]:
```

Doing `okta apps create jhipster` works. Looking at the cli code this seems correct (https://github.com/okta/okta-cli/blob/master/cli/src/main/java/com/okta/cli/commands/apps/templates/AppType.java) as jhipster is quick template.

### DIFF
--- a/pages/security.md
+++ b/pages/security.md
@@ -115,7 +115,7 @@ If you'd like to use Okta instead of Keycloak, it's pretty quick using the [Okta
 okta register
 ```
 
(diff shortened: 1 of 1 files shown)


mraible (approved):
I love it! You found a shortcut I didn't know about. If you run `okta apps create` and select **Web** for the first answer, you'll see **JHipster** in the 2nd question. `okta apps create jhipster` is much better!

Screen Shot 2021-01-16 at 1 11 32 PM

atomfrede:
So we could also change it to `okta apps create` and select `Web` then select `JHipster` but the shortcut is indeed better

PR #1067, opened 2021-01-23
jdl doc updates

deepu105:


pascalgrimaud:
@deepu105 : does it mean we can update these project too ?

https://github.com/jhipster/jhipster-oh-my-zsh-plugin/blob/main/jhipster.plugin.zsh#L9
https://github.com/jhipster/jhipster-fisher-plugin/blob/main/conf.d/jhipster.aliases.fish#L9

deepu105:
Yes @pascalgrimaud

PR #1071, opened 2021-01-24
Mongock is now used instead of Mongobee

murdos:
Mongock is now used instead of Mongobee since migration to Spring Boot 2.3

They have been really responsive to support Spring Data MongoDB 3 and MongoDB java driver 4.

Also added Mongock logo in server side stack


pascalgrimaud (approved):

pascalgrimaud:
Thanks @murdos !

PR #1090, opened 2021-03-04
JHipster App is deprecated

CGarces:
JHipster App is deprecated. I have changed the order of the items, using the terminal should be the preferred way.
Other possibility is just remove all references to JHipster App.


pascalgrimaud:
Thanks for your contribution @CGarces

PR #1099, opened 2021-03-11
Remove repeated 'a' article from problem-with-message.md

ramnasidharta:
[This page](https://www.jhipster.tech/problem/problem-with-message/) currently has a phrase starting with `JHipster is a an application [...]`, while it should be `JHipster is an application [...]`.


ruddell:
Thanks!

PR #1101, opened 2021-03-16
Update Okta instructions for new Admin Console

mraible:
While I was updating this page, I noticed we still reference UAA in it. Should we remove those docs, @pascalgrimaud?


pascalgrimaud:
@mraible : UAA should be removed here https://github.com/jhipster/jhipster.github.io/pull/1069 and this branch was merged into prepare-release-v7.0.0

PR #1103, opened 2021-03-18
Adding an additional note to configure corporate Outlook

jairo-henao:
When 'jhipster.mail.from' is not added or does not match 'spring.mail.username', the mail is not sent


pascalgrimaud (approved):

pascalgrimaud:
Thanks for your contribution @jairo-henao

PR #1107, opened 2021-03-20
New tip: How to remove 'Register account'

apuntandoanulo:
The generator adds by default the option to self-register. In some cases for security reasons it is required to delete it.


jdubois:
I'm merging and adding this to the index myself, thank you!

PR #1133, opened 2021-05-30
Add dropdown menu for JDL docs

iconben:
Current documents of JDL are not indexed in aside menu. This PR is to add dropdown menus under JDL so that navigation to JDL docs could be easier.

The pages that dropdown menus linked to are already existing pages, so nothing new, just links.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -215,10 +215,26 @@
                                     </div>
                                 </div>
                             </li>
-                            <li {% if '/jdl/' == page.url %}class="active" {% endif %}><a href="{{ site.url }}/jdl/"><span class="icon fa fa-fw fa-star"></span><span class="title">JDL</span></a></li>
                             <li class="panel panel-default dropdown">
-                                <a data-toggle="collapse" href="#dropdown-5"><span class="icon fa fa-fw fa-sitemap"></span><span class="title">Microservices</span></a>
-                                <div id="dropdown-5" class="panel-collapse collapse">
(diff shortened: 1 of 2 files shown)


_layouts/default.html
```diff
@@ -215,10 +215,26 @@
...
                             </li>
-                            <li {% if '/jdl/' == page.url %}class="active" {% endif %}><a href="{{ site.url }}/jdl/"><span class="icon fa fa-fw fa-star"></span><span class="title">JDL</span></a></li>
                             <li class="panel panel-default dropdown">
-                                <a data-toggle="collapse" href="#dropdown-5"><span class="icon fa fa-fw fa-sitemap"></span><span class="title">Microservices</span></a>
-                                <div id="dropdown-5" class="panel-collapse collapse">
+                                <a data-toggle="collapse" href="#dropdown-5"><span class="icon fa fa-fw fa-star"></span><span class="title">JDL</span></a>
+                                <div id="dropdown-5" class="panel-collapse collapse" style="height: 0px;">
+                                    <div class="panel-body">
+                                        <ul class="nav navbar-nav">
+                                            <li {% if '/jdl/intro' == page.url %}class="active" {% endif %}><a href="{{ site.url }}/jdl/intro"><i class="fa fa-fw fa-star"></i>Overview</a></li>
```
gmarziou, line 223: Rather than repeating same conditional construct, maybe you could use Jekyll includes with 3 params: page url, icon and text.

See https://jekyllrb.com/docs/includes/

iconben: It's good idea, thanks! Currently menu of other chapters are all in such straight plain html,  shall we merge this first and refactor later?

gmarziou: Sure.

PR #1134, opened 2021-05-31
Update enums.md

wanghengxn:
enum name is mandatory

### DIFF
--- a/pages/jdl/enums.md
+++ b/pages/jdl/enums.md
@@ -24,8 +24,8 @@ sitemap:
 Enumeration declaration is done as follows:
 
 ```
-enum [<enum name>] {
-  <ENUM KEY> ([<enum value>])
+enum <enum name> {
+  <ENUM KEY> [(<enum value>)]
 }
 ```
 


mshima (approved):

pages/jdl/enums.md
```diff
@@ -24,8 +24,8 @@ sitemap:
 Enumeration declaration is done as follows:
 
 ```
-enum [<enum name>] {
-  <ENUM KEY> ([<enum value>])
+enum <enum name> {
```
MathieuAA, line 27: thanks

pages/jdl/enums.md
```diff
@@ -24,8 +24,8 @@ sitemap:
...
+  <ENUM KEY> [(<enum value>)]
```
MathieuAA, line 28: I don't really see the point here

mshima: `([<enum value>])` implies that this is valid:
```
enum Foo {
  Bar ()
}
```
And this is invalid
```
enum Foo {
  Bar
}
```

MathieuAA: Ah very good point! Your eyes are better than mine

MathieuAA (approved):

PR #1136, opened 2021-05-31
Update relationships.md

wanghengxn:
The test results are as described above. 

jhipster -V
INFO! Using JHipster version installed globally
6.10.5

### DIFF
--- a/pages/jdl/relationships.md
+++ b/pages/jdl/relationships.md
@@ -126,9 +126,9 @@ relationship OneToOne {
   A{b} to B{a}
 }
 ```
-Not specifying an injected field is the short form of using having an bidirectional relationship.
+Not specifying an injected field is the short form of having a bidirectional relationship.
 
-another example:
+Another example:
 ```jdl
 relationship OneToOne {
   A{b} to B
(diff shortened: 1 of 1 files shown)


wanghengxn:
I found that there are too many things wrong with this document. Maybe you should find someone to check it out.

mshima (approved):

pages/jdl/relationships.md
```diff
@@ -122,12 +122,20 @@ relationship OneToOne {
 
 Note that this example is the same as:
 ```jdl
+relationship OneToOne {
+  A{b} to B{a}
+}
+```
+Not specifying an injected field is the short form of using having an bidirectional relationship.
+
+another example:
```
MathieuAA, line 131:
```suggestion
Another example:
```

pages/jdl/relationships.md
```diff
@@ -122,12 +122,20 @@ relationship OneToOne {
 
 Note that this example is the same as:
 ```jdl
+relationship OneToOne {
+  A{b} to B{a}
+}
+```
+Not specifying an injected field is the short form of using having an bidirectional relationship.
```
MathieuAA, line 129:
```suggestion
Not specifying an injected field is the short form of having a bidirectional relationship.
```

MathieuAA (changes requested):
Seems like some changes were made to the generation part and this doc hasn't been up to date... :(

PR #1137, opened 2021-06-01
fix hyperlink

patrickp-at-work:
You can’t use Markdown syntax inside block-level HTML tags. see https://www.markdownguide.org/basic-syntax/#html


pascalgrimaud:
Thanks @patrickp-at-work

PR #1140, opened 2021-06-08
Updated the deployment page and added missing options

MathieuAA:
Related to https://github.com/jhipster/generator-jhipster/issues/14817

### DIFF
--- a/pages/jdl/deployments.md
+++ b/pages/jdl/deployments.md
@@ -4,7 +4,7 @@ title: JHipster Domain Language - Deployments
 permalink: /jdl/deployments
 sitemap:
     priority: 0.5
-    lastmod: 2021-03-08T12:00:00-00:00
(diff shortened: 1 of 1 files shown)


pages/jdl/deployments.md
```diff
@@ -50,14 +50,14 @@ deployment {
...
 deployment {
   deploymentType docker-compose
   appsFolders [foo, bar]
   dockerRepositoryName "yourDockerLoginName"
 }
 
-// will be created under 'kubernetes' folder
+// will be created under the 'kubernetes' folder
 deployment {
   deploymentType kubernetes
```
mraible, line 62: Just curious - is it possible to change the name of the folder that's generated? For example, `k8s` instead of `kubernetes`?

MathieuAA: I don't know if there's an option for that, but it isn't possible (from what I've seen) with the JDL, yet.

PR #1145, opened 2021-06-25
Fixed lessOrEqualThan greaterOrEqualThan

volkovd:
Filter attributes lessOrEqualThan greaterOrEqualThan should be: lessThanOrEqual greaterThanOrEqual


pascalgrimaud:
Thanks for your contribution @volkovd

PR #1148, opened 2021-07-10
Fix broken links

GeraldScott:
Searched for broken links (404), fixed those where I could find updated web resources.


gmarziou:
Great work, thanks

PR #1149, opened 2021-07-12
chore: update aws.md added required permission

rgrativol:
Added AmazonS3FullAccess permission that is required to create the bucket when uploading the war file using the jhipster aws tool.


jdubois:
Thank you @rgrativol !

PR #1152, opened 2021-07-24
Add info about User entity restrictions

carestra:
Info about the user entity is not accessible in the microservice when using JWT or DTO, only when OAuth2 is use.


jdubois:
Thank you @carestra !

PR #1170, opened 2021-09-27
Fix proxy URL for npm-registry proxy call

swarajsaaj:
The API on marketplace details page to fetch package details has stopped working due to change in CORS proxy. It needs to include protocol now.
image

Maybe in future we can use netlify rewrites to have own proxy url instead of using an external one.

Proxy request via cors.bridged.cc changed to include protocol as
requirement

### DIFF
--- a/js/services.js
+++ b/js/services.js
@@ -53,7 +53,7 @@
             },
 
             getNpmInfo: function (npmPackageName) {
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Thanks @swarajsaaj !

PR #1171, opened 2021-10-09
add tips on generation and custom code combining

Tcharl:


Tcharl:
@DanielFran maybe?

pages/tips.md
```diff
@@ -16,26 +16,27 @@ This section contains **user-submitted tips'n tricks** on using JHipster.
...
-1. [Increase integration test performance by lazy bean initialization]({{ site.url }}/tips/027_tip_lazy_init_test_beans.html)
-1. [How to connect PGAdmin (PostgreSQL) to Heroku]({{ site.url }}/tips/028_tip_pgadmin_heroku.html)
-1. [Provide Internet Explorer support]({{ site.url }}/tips/028_tip_ie_support.html)
-1. [Improve developer experience if opening only front end in the IDE]({{ site.url }}/tips/029_tip_frontend_only.html)
-1. [Configure Redis leader follower(master-slave) replication]({{ site.url }}/tips/030_tip_redis_replication.html)
-1. [Running Protractor e2e tests within Intellij IDEA]({{ site.url }}/tips/031_tip_e2e_intellij.html)
-1. [Time Drift in Docker]({{ site.url }}/tips/032_tip_time_drift_docker.html)
-1. [User entity management]({{ site.url }}/user-entity/)
-1. [Remove the register account service]({{ site.url }}/tips/034_tip_remove_register_account_service.html)
+2. [Using Bootswatch themes]({{ site.url }}/tips/009_tips_using_bootswatch_themes.html)
```
DanielFran, line 19: @Tcharl Rollback those changes, not needed to increment the number. see https://www.jhipster.tech/tips/

DanielFran (changes requested):

pages/tips.md
Tcharl: Ok but why rollbacking? both works, and I prefer the second one, (and my and most of the IDE too, as they're automatically increasing the bit...)

crowne: The reason for keeping all the numbers labelled "- 1." is because it is less change, and a smaller diff will be stored by git.
If the numbers were stored in ascending order and someone needed to insert a new entry after #12 then all of the following numbers would need to be renumbered again causing a bigger diff to be stored by git.
Smaller diff's are easier to review to see where change has been made.

Tcharl:
Thanks Pascal!

pascalgrimaud:
Thank you @Tcharl :-)

PR #1187, opened 2022-01-23
Flag up that the User entity doesn't always exist

JavaDerek:
As microservices become more popular, we should let folks know on the page talking about the User entity that the rules for it change under microservices.

### DIFF
--- a/pages/user-entity.md
+++ b/pages/user-entity.md
@@ -9,6 +9,9 @@ sitemap:
 
 # <i class="fa fa-user"></i> The User entity
 
+NOTE: this is not created for microservices using some kinds of secvurity, as detailed
+<a href="/creating-microservices/">here</a>.
+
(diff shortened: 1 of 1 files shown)


pages/user-entity.md
```diff
@@ -9,6 +9,9 @@ sitemap:
 
 # <i class="fa fa-user"></i> The User entity
 
+NOTE: this is not created for microservices using some kinds of secvurity, as detailed
```
pascalgrimaud, line 12:
```suggestion
NOTE: this is not created for microservices using some kinds of security, as detailed
```

pascalgrimaud (changes requested):

pascalgrimaud (approved):

pascalgrimaud:
Thanks @JavaDerek

PR #1189, opened 2022-01-26
Replace ‘paginate D with pager’ on the last line with ‘paginate D wit…

omoniyi24:
…h pagination’ #17678


mraible:
If I use [JDL Studio](https://start.jhipster.tech/jdl-studio/), `paginate` is an option. Is this correct @MathieuAA?

MathieuAA:
If I'm not mistaken yes

PR #1198, opened 2022-03-02
Update documentation for generate-blueprint

mshima:
Closes https://github.com/jhipster/generator-jhipster/issues/16988.
Closes https://github.com/jhipster/generator-jhipster/issues/18304.
Closes https://github.com/jhipster/generator-jhipster/issues/14157.

To be merged once https://github.com/jhipster/generator-jhipster/pull/18306 is merged and released.

PR #1204, opened 2022-04-04
Add documentation for mobile apps and OIDC

mraible:


mraible:
I improved the headings by adding the following to the page so it renders a table of contents.

```
* Table of Contents
{:toc}
```

Screen Shot 2022-04-04 at 18 46 35

I removed the TOC before updating this PR. 

It might be nice to have H5s, but those render really small so I didn't use them.

mraible:
@deepu105 Do you know how to change the site so it doesn't redirect to the main site when viewing the preview link?

mshima:
> Do you know how to change the site so it doesn't redirect to the main site when viewing the preview link?

See the first file change: https://github.com/jhipster/jhipster.github.io/pull/1198/files
Need to remove `{{ site.url }}` from the links.

pages/security.md
```diff
@@ -183,6 +183,64 @@ heroku config:set \
   SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_OIDC_CLIENT_SECRET="$SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_OIDC_CLIENT_SECRET"
 ```
 
+#### Create a Native Okta App for Mobile
```
mshima, line 186: I get a bit confused with this sentence.
I keep thinking that I need an additional `Native App for Mobile`.
Would it be better to change to something like `Register a Native App for Mobile on Okta`?
Or `Create a Native App for Mobile on Okta` like bellow?

pages/security.md
```diff
@@ -278,6 +343,67 @@ You can put this in an `~/.auth0.env` file and run `source ~/.auth0.env` to over
 
 _Note_: If you're on `Windows`, you should install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10) so the `source` command will work.
 
+#### Create a Native Auth0 App for Mobile
```
mshima, line 346: Same as Okta.

pages/security.md
```diff
@@ -183,6 +183,64 @@ heroku config:set \
...
+Okta application configuration:
+Issuer:    https://dev-133337.okta.com/oauth2/default
+Client ID: 0oab8eb55Kb9jdMIr5d6
+```
+
+**NOTE**: You can also use the Okta Admin Console to create your app. See [Create a Native App](https://developer.okta.com/docs/guides/sign-into-mobile-app/create-okta-application/) for more information.
+
+#### Add Claims to Access Token
+
+In order to authentication successfully with your mobile app, you have to do a bit more configuration in Okta. Since the mobile client will only send an access token to JHipster, you need to 1) add a `groups` claim to the access token and 2) add a couple more claims so the user's name will be available in JHipster.
```
mshima, line 209:
```suggestion
In order to authenticate successfully with your mobile app, you have to do a bit more configuration in Okta. Since the mobile client will only send an access token to JHipster, you need to 1) add a `groups` claim to the access token and 2) add a couple more claims so the user's name will be available in JHipster.
```

pages/security.md
```diff
@@ -278,6 +343,67 @@ You can put this in an `~/.auth0.env` file and run `source ~/.auth0.env` to over
...
+    - Ionic: `http://localhost:8100`
+    - React Native: `http://localhost:19006`
+
+#### Update Your Ionic App
+
+Update `ionic/src/app/auth/auth-config.service.ts` to use the generated client ID:
+
+```ts
+environment.oidcConfig.server_host = this.authConfig.issuer;
+environment.oidcConfig.client_id = 'Dz7Oc9Zv9onjUBsdC55wReC4ifGMlA7G';
```
mshima, line 371:
```suggestion
environment.oidcConfig.client_id = '<your-client-id>';
```

mshima (approved):

pages/security.md
mraible: Fixed.

mraible:
> Need to remove `{{ site.url }}` from the links.

@mshima I don't think this is causing it. When I click on the [preview link](https://deploy-preview-1204--jhipster-site.netlify.app/security/) from Netlify in this PR, it redirects me to `https://www.jhipster.tech/security` rather than letting me view it. I tried adding conditional logic to the layout, but it didn't work, so I [removed it](https://github.com/jhipster/jhipster.github.io/pull/1204/commits/2c5dbebeafef1dc90e3d2b576fa1aa454aa9d033).

deepu105:
@mraible I think its coz of the redirect rules in the front matter

```
permalink: /security/
redirect_from:
  - /security.html
```

mraible:
@deepu105 I don't think that's it since it renders fine (and redirects properly) when running locally.

deepu105:
could it be something at netlify settings, that doesn't resolve the right base URL? or may be there is the hipster base URL hard coded somewhere?

deepu105:
We do have it hardcoded in the config, not sure how it works locally,  https://github.com/jhipster/jhipster.github.io/blob/main/_config.yml

Seems like a known issue https://answers.netlify.com/t/jekyll-site-uses-absolute-urls-deploy-previews-navigate-to-main-site/8441

mraible:
I'll merge this and open a new issue for the redirect/preview problem.

PR #1213, opened 2022-05-19
Update docker_compose.md

tonylook:
JIB is able to build the image starting from an image present in our local registry, it just need to explicitly say to pull this image from the local docker registry, then jib will transform it in jib format and cache it.


DanielFran (approved):

DanielFran:
Thanks @tonylook for your contribution

PR #1232, opened 2022-09-14
Remove expired discount codes

SoniaLahi:
This PR is about removing two discount codes that are displayed on the homepage but have expired a long time ago.
No big changes :ok_hand:


deepu105:
Thank you @SoniaLahi

PR #1248, opened 2022-11-30
Clarify caching strategy

CGarces:
This should be a question more than a PR because I'm not sure if is the correct behavior, please  merge with caution.

I'm configuring a CDN to serve and angular application generated with JHipster.

A lot of content from "/" are hashed (see the css in the image) so I think that is safe to cache the "/" content except the index.html

It's that correct?

image


mraible:
This looks correct to me. In reality, you could probably cache `index.html` since it doesn't change a whole lot. But I suppose it could be used to add new CSS or JS files between releases, so best not to cache.

mraible (approved):

atomfrede:
What about e.g. the `jhipster_family_member_1_head-192.png`? Sure it won't change but custom (static) resources will not be hashed (I think). So maybe we should cache `/` with caution?

mraible:
Yes, please do.

PR #1247, opened 2022-11-28
Update Elasticsearch docs for Heroku

mraible:

PR #1251, opened 2022-12-13
Elasticsearch documentation cleanup

saschaszott:
The support of embedded ES was dropped in JHipster. The documentation should reflect this.


mraible:
This has already been fixed in https://github.com/jhipster/jhipster.github.io/pull/1247. I'll try to get it merged today.

PR #1265, opened 2023-02-13
fix: update import gulp-imagemin

Jessy-BAER:
Fix issue https://github.com/jhipster/generator-jhipster/issues/21119


DanielFran:
Thanks @Jessy-BAER for the contribution

PR #1273, opened 2023-04-02
Add mention of tool to transform SQL to JDL

Blackdread:
Fixes https://github.com/jhipster/generator-jhipster/issues/20759

Add mention of tool to transform SQL to JDL

@pascalgrimaud Is `main` or `master` branch that should be used? It seems that `main` should actually be the branch to use (Previous PR https://github.com/jhipster/jhipster.github.io/pull/1271 was merged in master but main and master now differs for quite some time)


pascalgrimaud:
@Blackdread : it's the main branch, which is the one which is used at JHipster.tech

PR #1301, opened 2023-08-19
consul as default

deepu105:
Related to https://github.com/jhipster/generator-jhipster/issues/19109


DanielFran (approved):

pascalgrimaud:
@deepu105 @DanielFran : I wonder if we should merge this, before the next release? As the current site will be desync with the last release

DanielFran:
@pascalgrimaud Dev in generator is part already of 1st beta, so it was missing only some docs to oficialize it

PR #1323, opened 2023-11-08
cleanup

mshima:


mshima:
Should gcloud, openshift and others deployments be removed or add a warning that the generator have been removed?

DanielFran:
> Should gcloud, openshift and others deployments be removed or add a warning that the generator have been removed?

Better to remove

pascalgrimaud:
Not sure about removing these pages, as it's not good for SEO
That's why there are still other old pages

mraible:
I'm not sure we should be concerned about SEO. It's not like we have a lot of competition for terms like "JHipster GAE." If we do, I think that's a good thing.

atomfrede:
:facepalm: I should have looked at the existing PRs before doing https://github.com/jhipster/jhipster.github.io/pull/1328

atomfrede (approved):

atomfrede:
Let's merge this first in my other PR the heroku part is the most important thing.

PR #1338, opened 2024-03-21
Update relationships.md

Lalithkumarponnambalam:
### **Update relationships.md content**
The following issue will be rectified:

- jhipster/generator-jhipster/issues/25584

I have referenced the "joint fields" from the JHipster Pro documentation. The correct link for the documentation is:

- [JHipster Pro Documentation - Relationships with Joint Fields](http://www.jhipster.pro/docs/jdl/relationships.en#with-joint-fields)

### DIFF
--- a/pages/jdl/relationships.md
+++ b/pages/jdl/relationships.md
@@ -157,6 +157,22 @@ entity.
 
 ---
 
+#### With joint fields
+
+Used to specify which column in the entities are used for the join (default to id)
+
+```jdl
+relationship OneToOne {
+  A{b(name)} to B{a(name)}
+}
+```
+
+It roughly translates to `SQL: JOIN B b with a.name = b.name`
+
+---
+
+---
+
 #### With methods
 
 ```jdl


OmarHawk:
Is this correct? To me the described syntax rather means "show name in any referencing attribute, like dropdown or list view link", the join will still always be executed based on the foreign key relationship afaik.

Lalithkumarponnambalam:
@OmarHawk  yes this is correct. In the JHipster Pro Documentation, there is an explanation of join fields in relationships, which can be found at the following link: [JHipster Pro Documentation - Relationships with Joint Fields](http://www.jhipster.pro/docs/jdl/relationships.en#with-joint-fields).

OmarHawk:
Still not convinced. We are using the syntax and it works as **I did** describe. See also the syntax description here: https://www.jhipster.tech/jdl/relationships#syntax ---> `<display field>`

Also the .pro site looks kind of unofficial and rather outdated.

mshima:
That’s not correct.
```
relationship OneToOne {
  A{b(name)} to B{a(name)}
}
```

means that when showing A, B will have `b.name` field as label (instead of b.id) in the frontend. Same for B.
joins always use ids.

mraible:
We updated the website with a new design. Can you please update this issue accordingly?

mraible:
Closing due to lack of response.

PR #1347, opened 2024-05-08
Add globally installed jhipster alternative for upgrade

mshima:
`npx generator-jhipster@latest upgrade` doesn’t work as expected.
See https://github.com/npm/cli/issues/7486

### DIFF
--- a/pages/upgrading_an_app.md
+++ b/pages/upgrading_an_app.md
@@ -39,28 +39,15 @@ Go into the application's root directory:
 
 To upgrade your application, type:
 
-`npx generator-jhipster@latest upgrade`
+```
+npm install -g generator-jhipster@latest
+jhipster upgrade
+```
 
 Here are the options you can pass:
 
-* `--verbose` - Log each step of the upgrade process in detail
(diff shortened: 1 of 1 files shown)


mraible (approved):

mshima:
@mraible can you reproduce https://github.com/npm/cli/issues/7486?

mraible:
@mshima If I try `npx generator-jhipster@8.4.0 --version`, I get `8.4.0` as an output. This is with nothing installed globally. If I install 8.3.0 first, it works too.

Screenshot 2024-05-08 at 12 35 37 PM

pages/upgrading_an_app.md
```diff
@@ -41,27 +41,18 @@ To upgrade your application, type:
 
 `npx generator-jhipster@latest upgrade`
 
+An alternative is to use global instalation:
```
mraible, line 44:
```suggestion
An alternative is to use global installation:
```

mshima:
https://github.com/npm/cli/issues/7486 is marked as a valid bug.
It’s not too bad. So I think we can keep the npx upgrade command.

PR #1348, opened 2024-05-12
Replace Twitter by X

qmonmert:
@mraible @jdubois @pascalgrimaud @DanielFran are you ok with this update?

I didn't update Twitter logo because the X Twitter is not available with font-awesome 4, we need to upgrade to 6 (with a lot of work).

There are only two logos:
image
image

### DIFF
--- a/_posts/2013-12-06-jhipster-release-0.3.0.md
+++ b/_posts/2013-12-06-jhipster-release-0.3.0.md
@@ -43,5 +43,5 @@ Help and bugs
 
 If you find any issue with this release, don't hesitate to:
 
-- Contact [@jhipster](https://twitter.com/jhipster) on Twitter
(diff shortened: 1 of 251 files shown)


mraible:
I still prefer to call it Twitter.

pascalgrimaud:
I don't think we need to update old release notes
We could update the current main pages, and that's all

qmonmert:
@pascalgrimaud these 4 files or just `index.html`?
image

pascalgrimaud:
Yes I think @qmonmert

qmonmert:
@pascalgrimaud I updated with this solution.

@mraible WDYT about this?

mraible:
@qmonmert I don't feel strongly enough about it to approve or deny. ;)

qmonmert:
Ok I close this PR

PR #1350, opened 2024-05-26
French Version Of Jhipster Website

maximeetundi:


pascalgrimaud:
Thanks for proposing this, but it should be discussed first before you started the work.

The main problem to propose a translated version is it won't be maintained after some weeks.
We can already see the obsolete version, which are not maintained at all :
- https://www.jhipster.tech/cn/team/
- https://www.jhipster.tech/jp/team/

I think these 2 versions should be deleted -> cc @PierreBesson

PR #1379, opened 2024-09-06
Fix broken links to documentation archive

Alchemik:
Image

### DIFF
--- a/docs/customizing-bootstrap-3.mdx
+++ b/docs/customizing-bootstrap-3.mdx
@@ -9,4 +9,4 @@ As of JHipster 5, AngularJS and Bootstrap 3 are not supported anymore with JHips
 
 We have great support for Angular and React, which are both more modern tools, we hope you will enjoy them!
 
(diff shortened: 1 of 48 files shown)


mraible:
Why is this necessary? If I go to https://www.jhipster.tech/documentation-archive (without the slash), it redirects to https://www.jhipster.tech/documentation-archive/ (with the slash).

Alchemik:
Sorry, I should clearly state the issue in the PR description. I stumbled upon this problem on every page which contains link to the archive - as you can on the attached gif, when the user clicks on the link leading to documentation archive, the redirection does not work as intended and "Page not found" is presented instead. For example, try to access archive from the page [https://www.jhipster.tech/2018/05/03/jhipster-release-5.0.0-beta.1.html#documentation](https://www.jhipster.tech/2018/05/03/jhipster-release-5.0.0-beta.1.html#documentation)

SvMak:
Thanks a lot for your bug report. Description about this issue and solution in #1380

PR #1397, opened 2024-10-10
Add microfrontends example to documentation

mraible:


mshima:
I think we can link to https://module-federation.io/.
We probably will switch to `module federation 2.0` plugins in the future.

mraible:
@mshima Updated to link to https://module-federation.io/.

mshima (approved):

PR #1416, opened 2024-11-06
Add a language switcher

hide212131:
Related to https://github.com/jhipster/jp/issues/36

This is a proposal for a feature allowing the selection of multiple languages on the site.
The envisioned layout for the English site is as follows:

image

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -41,7 +41,15 @@ const config: Config = {
   // may want to replace "en" with "zh-Hans".
   i18n: {
     defaultLocale: 'en',
-    locales: ['en'],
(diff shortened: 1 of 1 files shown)


mraible:
I'm not sure how this helps if there isn't a Japanese translation of the site.

hide212131:
@mraible The Japanese site has been around for a while. I just updated it to the new site.
https://www.jhipster.tech/jp/
[The repositories are separate](https://github.com/jhipster/jp), but by merging this PR, it will be possible to navigate to the Japanese site. The link in the opposite direction has also been implemented.

mraible (approved):

PR #1417, opened 2024-11-07
Reduce the size of the build folder

SvMak:
- part of images converted to webp format;
- remove unused images.

This should reduce the size of the build folder.

Related to: https://github.com/jhipster/generator-jhipster/issues/27770


mraible:
Can you add some stats on repo size before and after this change?

SvMak:
This mainly concerns the build folder, since this is what we add to our archive.

Screenshot from 2024-11-07 21-19-40

Screenshot from 2024-11-07 21-22-48

SvMak:
> This mainly concerns the build folder, since this is what we add to our archive.
> 
> Screenshot from 2024-11-07 21-19-40
> 
> Screenshot from 2024-11-07 21-22-48

@mraible What do you think?

mraible:
If the images are converted from png to webp, does the quality decline? I remember that being a problem last time.

SvMak:
> If the images are converted from png to webp, does the quality decline? I remember that being a problem last time.

Potentially yes, I used a factor of 0.75 for compression, but I didn't notice a big difference.

I would be happy if you could compare the quality of the images on your macbook.

PR #1479, opened 2025-04-08
Fixed update for react 19

SvMak:
- added .npmrc with legacy-peer-deps;
- updated dependencies.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+legacy-peer-deps=true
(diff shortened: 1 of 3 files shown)


mraible (approved):

mshima:
I don't think `legacy-peer-deps` is required.
react/react-dom should be updated together.

SvMak:
Ok, I'll close this.

